PR #33, opened 2022-02-10
chore: allow for relative dependency paths

EverlastingBugstopper:
in `federation` integration tests we overwrite the value of the dependency in package.json. this pr allows us to detect that and resolve its version in the same way so all the other packages can be updated in the build script.


EverlastingBugstopper:
punting this

PR #34, opened 2022-02-11
remove mkdir -p  in the build step.

o0Ignition0o:
Windows doesn't support the -p flag, which is only useful on linux if parents need to be created. It turns out we don't need this here anyway, because we're creating two directories from cwd.

What happens and seems to fail the CI is that windows will error out when invoking `mkdir` if the target directory already exists. So there are two issues in the current build script on windows:

- mkdir creates three directories: namely `-p` `bundled` and `js-dist`
- mkdir errors out if a directory already exists, which happens to be the case in the `apollographql/router` CI because we have it in our cache.

In this pull request, we run `make-dir bundled js-dist && ...` so that we don't catch any `directory already exists` error, which behaves as expected on windows and linux / darwin.

### DIFF
--- a/router-bridge/package.json
+++ b/router-bridge/package.json
@@ -36,6 +36,7 @@
     "buffer": "^6.0.3",
     "console-polyfill": "^0.3.0",
     "esbuild": "^0.14.20",
+    "make-dir-cli": "^3.0.0",
     "path-browserify": "^1.0.1",
     "prettier": "^2.5.1",
     "typescript": "^4.5.5",
--- a/router-bridge/package.json
+++ b/router-bridge/package.json
@@ -4,7 +4,7 @@
   "version": "2.0.0-alpha.5",
   "description": "Apollo Router JS Bridge Entrypoint",
   "scripts": {
-    "build": "mkdir -p bundled js-dist && tsc && node esbuild/bundler.js",
+    "build": "mkdir bundled; mkdir js-dist; tsc && node esbuild/bundler.js",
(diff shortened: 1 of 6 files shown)


o0Ignition0o:
This is currently a draft because I would like to see why the CI hasn't spotted it, and if any changes need to be made in xtask or .circleci/config.yml

router-bridge/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "2.0.0-alpha.5",
   "description": "Apollo Router JS Bridge Entrypoint",
   "scripts": {
-    "build": "mkdir -p bundled js-dist && tsc && node esbuild/bundler.js",
```
abernix, line 7: One of the things that `-p` offers is that it will not error when the directory already exists.  We may want that to be maintained for local development reasons?  Maybe we should just use [`make-dir-cli`](https://www.npmjs.com/package/make-dir-cli) or put [`make-dir`](https://npm.im/make-dir) into `esbuild/bundler.js`?

o0Ignition0o: Indeed! I was wondering to what extent we would want to rely on "more tools", but it definitely looks the correct way to go

> I mention this is a workaround because I feel like depending on how complex the build script becomes, we might want to leverage [cross-env](https://www.npmjs.com/package/cross-env) or other packages to run OS specific commands

Let's give this a try

o0Ignition0o:
Ok I think the last remaining step is to remove tsconfig.buildinfo before we `save_cache` in ci

EverlastingBugstopper (approved):
lgtm 😀

PR #36, opened 2022-02-14
chore: add `assertion_lines` to `insta` snaps

abernix:
Per the new version of `insta`:

    Insta now memorizes assertion line numbers in snapshots. While these will
    quickly be old, they are often useful when reviewing snapshots immediately
    after creation with cargo-insta.

Ref: https://github.com/mitsuhiko/insta/blob/e8f3f2782e24b4eb5f256f94bbd98048d4a716ba/CHANGELOG.md#180
Ref: https://github.com/mitsuhiko/insta/blob/966320e81912b3f92407a0ff8991ba7b6233c9c9/src/output.rs#L64


EverlastingBugstopper (approved):
Sure

Issue #25, opened 2022-02-09
router-bridge: Re-introduce `validate` on the `main` (v2) branch


abernix:
When I initially switched to using the new `query-planner`, via https://github.com/apollographql/federation/commit/9cb220135d776043acfe0cee411277a8f2ce0976, I removed our invocation of `validate` on the premise that we did all the validations inside of the Federation's `buildQueryPlan` and `operationFromDocument`.   While we do a _lot_ in those, we do not do the complete suite of validations and we should add that back.

This can be seen rather clearly in the diff on the above commit, but roughly this consists of:

1. **Adding back**
	```js
	  const validationErrors = validate(composedSchema, query);
	  if (validationErrors.length > 0) {
	    return { errors: validationErrors };
	  }
	```
2. Making sure the Router itself can recognize that as a validation error and thread that through to [this code](https://github.com/apollographql/router/pull/309#discussion_r802030293) (see https://github.com/apollographql/router/pull/309)
3. Adding a test to the Router to make sure that we test for validation to have happened.

abernix:
Ok, this is unblocked.

PR #37, opened 2022-02-14
router-bridge: Introduce native `graphql` (JS) `validate` to the `plan`

abernix:
While Federation's `buildQueryPlan` function does a fair bit of validation in order to properly build our plan (and leverages `graphql`'s [`validate`] function do most of that), it *does not* do all of the [default validations] that are specified.

To be thorough, we must call `validate`, as we did [prior] to updating to the new Federation v2 alpha query planner.

This also introduces a number of tests which exercise specific validations which we know as of today are not part of federation:

 - `invalid_graphql_validation_1_is_caught`: [`NoFragmentCyclesRule`](https://github.com/graphql/graphql-js/blob/95dac43fd4bff037e06adaa7cfb44f497bca94a7/src/validation/rules/NoFragmentCyclesRule.ts)
 - `invalid_graphql_validation_2_is_caught`: [`ScalarLeafsRule`](https://github.com/graphql/graphql-js/blob/95dac43fd4bff037e06adaa7cfb44f497bca94a7/src/validation/rules/ScalarLeafsRule.ts)
 - `invalid_graphql_validation_3_is_caught`: [`NoUnusedFragmentsRule`](https://github.com/graphql/graphql-js/blob/95dac43fd4bff037e06adaa7cfb44f497bca94a7/src/validation/rules/NoUnusedFragmentsRule.ts)

I chose three random but different types to make sure we're capturing it.

[prior]: https://github.com/apollographql/federation/commit/9cb220135d776043acfe0cee411277a8f2ce0976
[`validate`]: https://github.com/graphql/graphql-js/blob/95dac43fd4bff037e06adaa7cfb44f497bca94a7/src/validation/validate.ts#L38
[default validations]: https://github.com/graphql/graphql-js/blob/95dac43fd4bff037e06adaa7cfb44f497bca94a7/src/validation/specifiedRules.ts#L76-L103

Closes: https://github.com/apollographql/federation-rs/issues/25

### DIFF
--- a/router-bridge/src/plan.rs
+++ b/router-bridge/src/plan.rs
@@ -186,6 +186,142 @@ mod tests {
     }
 
     #[test]
+    // A series of queries that should fail graphql-js's validate function.  The federation
+    // query planning logic automatically does some validation in order to do its duties.
+    // Some, but not all, of that validation is also handled by the graphql-js validator.
+    // However, we are trying to assert that we are testing graphql-js validation, not
+    // Federation's query planner validation.  So we run a few validations which we do not
+    // expect to every show up in Federation's query planner validation.
+    // This one is for the NoFragmentCyclesRule in graphql/validate
+    fn invalid_graphql_validation_1_is_caught() {
+        let result = Err::<serde_json::Value, _>(PlanningErrors {
+            errors: vec![PlanningError {
+                message: Some("Cannot spread fragment \"thatUserFragment1\" within itself via \"thatUserFragment2\".".to_string()),
+                extensions: None,
+            }],
+        });
+
+        assert_eq!(
+            result,
+            plan(
+                OperationalContext {
+                    schema: SCHEMA.to_string(),
+                    // These two fragments will spread themselves into a cycle, which is invalid per NoFragmentCyclesRule.
+                    query: "\
+                    fragment thatUserFragment1 on User {
+                        id
+                        ...thatUserFragment2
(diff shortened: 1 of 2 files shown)


router-bridge/src/plan.rs
```diff
@@ -225,7 +361,7 @@ mod tests {
     }
 
     #[test]
-    fn invalid_query_is_caught() {
+    fn syntactically_incorrect_query_is_caught() {
```
abernix, line 364: I renamed this just to add clarity — since this is a parsing error, rather than a validation error.

o0Ignition0o (approved):
Looks good to me!

I feel like some of the validation steps can occur before the actual query pipeline, eg when updating the supergraph schema. 

Would it be worth opening an issue on the router, to add schema validation capabilities that would allow us to refuse to update the router with an invalid schema?

abernix:
@o0Ignition0o That's a prudent observation, and you're correct.  We could use [`validateSDL`](https://github.com/graphql/graphql-js/blob/95dac43fd4bff037e06adaa7cfb44f497bca94a7/src/validation/validate.ts#L91-L108) to do some validations on the schema (which ultimately happen as part of `validate`, but we could do one better to avoid that schema from going into operation in the Router). (Though practically speaking, I believe our incoming schemas will typically be validated as part of Studio + Composition ahead of time).

I'll open an issue.

PR #41, opened 2022-02-22
fix the observed path for rerun-if-changed

Geal:
if the path does not exist, the build.rs task will rerun every time


Geal:
thanks!

Issue #42, opened 2022-02-01
No actionable error message on locally missing subgraph type (`supergraph compose`)


dchambers:
`supergraph compose` fails with no actionable error messages when a subgraph depends on a type that it does not define, **but** which some other subgraph does.

For example, given `a.graphql`:

```gql
extend type Query {
  fieldA: Uri
}
```

and `b.graphql`:

```gql
extend type Query {
  fieldB: Uri
}

scalar Uri
```

and a `supergraph.yml` config file pointing at these files:

```yml
subgraphs:
  films:
    routing_url: https://a.example.com
    schema:
      file: ./a.graphql
  people:
    routing_url: https://b.example.com
    schema:
      file: ./b.graphql
```

then we get:

```sh
rover supergraph compose --config supergraph.yml
error[E029]: Encountered 1 build error while trying to build a supergraph.

Caused by:
    Something went wrong! No build errors were recorded, but we also build a valid supergraph SDL.
    
        The subgraph schemas you provided are incompatible with each other. See https://www.apollographql.com/docs/federation/errors/ for more information on resolving build errors.
```

This leads to the end-developer needing to perform a binary search, which is hard with an SDL because you have to be careful not to remove types that are depended on by other types, otherwise you get completely different errors.

EverlastingBugstopper:
This is (mostly) fixed in Rover 0.4.5. Output looks like this for fed 1 and fed 2 respectively.

fed1 output is still a bit janky, i'm thinking we'll need to update the harmonizer-js lib to spit out `[{message: e.toString()}]` rather than just `[err]` for those errors that say "an unknown error occurred during the build". i've opened #70 to further fix this up.

```
$ rover supergraph compose --config ./examples/federation-rs/issue-42/supergraph.yaml     
error[E029]: Encountered 6 build errors while trying to build a supergraph.

Caused by:
    Encountered 6 build errors while trying to build the supergraph.
    UNKNOWN: An unknown error occurred during the build.
    UNKNOWN: An unknown error occurred during the build.
    UNKNOWN: Unknown type "Uri".
    UNKNOWN: Field "Query.fieldA" can only be defined once.
    UNKNOWN: Unknown type "Uri".
    UNKNOWN: Type Query must define one or more fields.
    
        The subgraph schemas you provided are incompatible with each other. See https://www.apollographql.com/docs/federation/errors/ for more information on resolving build errors.
```

```
[...]

PR #47, opened 2022-03-02
Api schema support

### LINKED ISSUE #46: Changes to support API schema
The following changes need to be made to support API schemas:

* Given a schema generate the API schema.
* Validation for before query planning should use API schema. 

Returning of API schema needs to be a separate call to query planning as we will need to return the API schema for introspection.

BrynCooke:
This PR updates existing query planning and introspection to use API schema and and also adds a dedicated way for rust to generate the API schema.

Fixes #46

### DIFF
--- a/router-bridge/js-src/introspection.ts
+++ b/router-bridge/js-src/introspection.ts
@@ -1,10 +1,12 @@
 import {
-  buildSchema,
+  buildSchema as gqlBuildSchema,
   ExecutionResult,
   GraphQLError,
   GraphQLSchema,
   graphqlSync,
 } from "graphql";
+
+import { buildSchema } from "@apollo/federation-internals";
 
 export function batchIntrospect(
   sdl: string,
@@ -12,7 +14,13 @@ export function batchIntrospect(
 ): ExecutionResult[] {
   let schema: GraphQLSchema;
   try {
-    schema = buildSchema(sdl);
+    // First go through regular schema parsing
+    gqlBuildSchema(sdl);
+
+    // Now try to get the API schema
+    let composedSchema = buildSchema(sdl);
+    let apiSchema = composedSchema.toAPISchema();
+    schema = apiSchema.toGraphQLJSSchema();
   } catch (e) {
     return Array(queries.length).fill({
       errors: [e],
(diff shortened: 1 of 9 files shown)


router-bridge/js-src/introspection.ts
```diff
@@ -1,18 +1,26 @@
 import {
-  buildSchema,
+  buildSchema as gqlBuildSchema,
   ExecutionResult,
   GraphQLError,
   GraphQLSchema,
   graphqlSync,
 } from "graphql";
 
+import { buildSchema } from "@apollo/federation-internals";
```
EverlastingBugstopper, line 9: small unnecessary nit: since we rename the one buildSchema might be worth renaming the other as well

```suggestion
import { buildSchema as buildFederatedSchema } from "@apollo/federation-internals";
```

router-bridge/src/api_schema.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// An error which occurred during JavaScript api schema generation.
+///
+/// The shape of this error is meant to mimick that of the error created within
+/// JavaScript, which is a [`GraphQLError`] from the [`graphql-js`] library.
+///
+/// [`graphql-js']: https://npm.im/graphql
+/// [`GraphQLError`]: https://github.com/graphql/graphql-js/blob/3869211/src/error/GraphQLError.js#L18-L75
+#[derive(Debug, Error, Serialize, Deserialize, PartialEq, Clone)]
+pub struct ApiSchemaError {
```
EverlastingBugstopper, line 19: doesn't have to be part of this PR but we use a similar error type across most of these packages. i wonder if we might consider writing some higher level type in `apollo-federation-types` where improvements would be delivered to all of the sub-packages. for instance, i imagine that there's some sort of case where this struct doesn't quite deserialize things properly (though I haven't investigated the root cause as of yet) https://github.com/apollographql/federation-rs/issues/42

EverlastingBugstopper (approved):

router-bridge/js-src/introspection.ts
BrynCooke: Fixed

router-bridge/src/api_schema.rs
BrynCooke: Agree, but let's keep refactor later. This may become redundant soon.

BrynCooke:
Going for merge. It looks like it's working correctly from the router side.

PR #65, opened 2022-03-11
fix: generate deno snapshots in build.rs

o0Ignition0o:
In #58 I added a snapshots generation phase to our harmonizer packages.

However I created a bug because the snapshot generation phase would happen when the harmonizer-0 and harmonizer-2 packages are generated.

This means downstream users who would rely on published harmonizer crates would get snapshots generated by an other computer / architecture.

This triggered a bug that would fail any harmonizer call when relying on the published crate.

This commit adds a build.rs script to the published crates, so that snapshots are created when downstream binaries are built.


garypen (approved):

harmonizer-2/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/apollographql/federation/"
 license-file = "./LICENSE"
 readme = "README.md"
 include = [ "dist/**/*", "deno/**/*", "src/**/*", "snapshots/**/*", "Cargo.toml" ]
+exclude = [ "snapshots/query_runtime.snap" ]
```
EverlastingBugstopper, line 13: should likely exclude `snapshots/**/*` from the `include` here, i added it while you weren't looking bc i didn't realize they were architecture dependent

PR #74, opened 2022-03-15
fix: npm install instead of npm ci

EverlastingBugstopper:
if you take a look at the diff you'll see that `package.json` is not updated and the `package-lock.json` is updated - meaning our last publishes contained incorrect versions because the lockfile was out of date.

we need to be running `npm install` instead of `npm ci` in debug mode because `npm ci` doesn't update the `package-lock.json`, and when we're running locally we want that to happen. in release mode we _don't_ want to update `package-lock.json` so that what we publish reflects what's in the lockfile.

### DIFF
--- a/build_harmonizer.rs
+++ b/build_harmonizer.rs
@@ -25,21 +25,34 @@ fn main() {
     create_snapshot().expect("unable to create v8 snapshot: query_runtime.snap");
 }
 
-// runs `npm ci` && `npm run build` in the current `harmonizer-x` workspace crate
+// runs `npm install` && `npm run build` in the current `harmonizer-x` workspace crate
 fn bundle_for_deno() {
     let npm = which::which("npm").expect("You must have npm installed to build this crate.");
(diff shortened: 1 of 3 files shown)


build_harmonizer.rs
```diff
@@ -25,21 +25,34 @@ fn main() {
...
+    } else {
+        println!(
+            "cargo:warning=running `npm ci` in {}",
+            &current_dir.display()
+        );
+        assert!(Command::new(&npm)
+            .current_dir(&current_dir)
+            .args(&["ci"])
+            .status()
+            .expect("Could not get status of `npm install`")
```
abernix, line 53: Not sure if this intentional or not.

```suggestion
            .expect("Could not get status of `npm ci`")
```

abernix (approved):

build_harmonizer.rs
EverlastingBugstopper: good catch ty

PR #79, opened 2022-03-16
chore: delete bespoke 'stage' implementation in favor of multiple workspaces

EverlastingBugstopper:
previously, you had to run `cargo xtask prep --package composition@v2.0.0-preview.7 --stage stage-02 && cargo xtask prep --package composition@v0.35.3 --stage stage-0` in order to get local versions of what would really get published. this was slow, not obvious, and bespoke.

now, there are 3 separate cargo workspaces

a root  workspace containing `apollo-federation-types` and `xtask`, a `federation-1` workspace containing v0.x of `harmonizer` and `supergraph`, and a `federation-2` workspace containing `v2.x` of `harmonizer`, `supergraph`, and `router-bridge`.

there is also a vs code workspace that makes sure rust-analyzer is picking up all of the workspaces it should, there is likely room for improvement here.

---

now, you can refer to individual packages by `git` tag and by `path` without doing any preparation. the repo looks like it should and there are no weird version negotiations with `Cargo.publish.toml`s scattered around.

`cargo xtask lint` and `cargo xtask test` now run three separate `cargo` processes, one for each workspace, and `cargo xtask prep` is now removed since we don't need to do _anything_ there that we were doing previously.


EverlastingBugstopper:
I've made a PR in Rover that proves that you can now build the versions of harmonizer you need by referring to the git repository.  including multiple versions at the same time: https://github.com/apollographql/rover/pull/1045

this was not possible before!

abernix:
nice!

PR #84, opened 2022-03-28
Add sha256sums.txt as a required build artifact

benweatherman:
sha256sums.txt is built during the release process, but it was never added to the list of artifacts we look for during publish. Now publishing will correctly check for sha256sums.txt.


EverlastingBugstopper (approved):
🔥 you da bomb

PR #87, opened 2022-03-28
fix: fixes crates.io publish for harmonizer

EverlastingBugstopper:
dev builds should work the same (except now we're putting generated snapshots into `$OUT_DIR` instead), and when publishing we don't run `npm install` or anything like that, we just take the latest bundled stuff (which should already have been built) and we _only_ generate the snapshot. this means we dont need info from package.json or to download anything from npm to build the crate if we're pulling from crates.io 😄 if you refer to this crate by path or by git dependency, you'll still need npm in order to build, but otherwise you won't! the router team may want to start publishing router-bridge to crates.io and referring to it by version in their Cargo.toml to take advantage of this (and remove npm installs from their CI pipeline)

### DIFF
--- a/federation-1/harmonizer/build.rs
+++ b/federation-1/harmonizer/build.rs
@@ -1,1 +1,0 @@
-../../build_harmonizer.rs
--- a/federation-1/harmonizer/build.rs
+++ b/federation-1/harmonizer/build.rs
@@ -1,0 +1,1 @@
+../../build_harmonizer.rs
--- a/federation-1/harmonizer/build.rs
+++ b/federation-1/harmonizer/build.rs
@@ -21,7 +21,8 @@ fn main() {
     let current_dir = std::env::current_dir().unwrap();
 
     // only do `npm` related stuff if we're _not_ publishing to crates.io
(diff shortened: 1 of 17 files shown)


federation-1/harmonizer/build.rs
```diff
@@ -10,25 +11,29 @@ use toml_edit::{value as new_toml_value, Document as TomlDocument};
...
-    bundle_for_deno();
-    // we still need to create deno snapshots in order to run tests locally
-    // the actual deno snapshot creation for downstream users will happen in the respective
-    // harmonizer-0 and harmonizer-2 build.rs file
-    create_snapshot().expect("unable to create v8 snapshot: query_runtime.snap");
+    let current_dir = std::env::current_dir().unwrap();
+
+    // only do `npm` related stuff if we're _not_ publishing to crates.io
+    // package.json is not in the `includes` section of `Cargo.toml`
+    if std::fs::metadata("./package.json").is_ok() {
```
o0Ignition0o, line 25: that's elegant! 🎉

o0Ignition0o (approved):

PR #90, opened 2022-03-31
feat: Bump deno version + refactor the query planner.

o0Ignition0o:
feat: Bump deno version + refactor the query planner.

This Pull request bumps router-bridge deno to version 0.124.0 (which is the latest deno_core version that is compatible with once_cell v1.9.0, which is required by harmonizer's deno).

This bump introduces the `#[op]` macro which is the official supported way to declare / expose rust functions to the deno runtime.

The bump also introduces a new way to register resources (such as channels) to the state, which allowed us to refactor the query planner:
Instead of spinning up a new deno runtime everytime we need a plan to happen, we have a runtime that is always available, and we send requests to it via async message passing capabilities.

This allows us to:
- Have only one runtime running (thus improving the overall predictability / memory footprint of the bridge)
- Validate the schema only once at QueryPlanner instanciation, (thus making the instanciation fallible _and_ avoid performing schema validation every time we would like to run query planning
-  Plan a query against an already instanciated QueryPlanner (thus improving the average query planning time from ~60ms to ~400us. [yes, microseconds](https://github.com/apollographql/router/pull/768#issuecomment-1085703557).)
 
While the changes here focus on router-bridge only for now, we could bump the deno runtime on the other crates here as well. I'd be happy to help in this respect.
In the meantime both deno_versions have subtle different v8 requirements, which prompted me to make router-bridge an optional member of the federation-2 workspace, so it can be compiled separately.


Geal (approved):

federation-2/router-bridge/js-src/runtime.js
```diff
@@ -1,17 +1,13 @@
...
-// This maps op names to their id's.
-Deno.core.ops();
-
 // Then we define a print function that uses
 // our op_print op to display the stringified argument.
 const _newline = new Uint8Array([10]);
 
 function print(value) {
-  Deno.core.dispatchByName("op_print", 0, value.toString(), _newline);
+  Deno.core.print(value.toString(), _newline);
```
o0Ignition0o, line 6: now that this is in place, we can definitely expose log functions that would leverage tracing eg log.warn() log.error() log.debug() etc 🎉

EverlastingBugstopper (approved):

PR #91, opened 2022-04-01
feat: adds federation_version to supergraph config

EverlastingBugstopper:
this allows folks to specify `federation_version` in their supergraph.yaml. if it's 0 or 1 it will be `FederationOne`, if it's 2 it'll be `FederationTwo`. if it's blank, it'll default to federation 2.

the logic for determining which binary to run will be in rover, the supergraph binaries themselves will error if it is being used with a supergraph.yaml that doesn't have the right version. we may in the future support a full `semver::VersionReq`, but this is a good start for now, and making it a `VersionReq` would be backwards compatible down the line.

i also added a `./install.sh` that will build and move the proper binaries to `~/.rover/bin` for proper testing 😄


federation-1/supergraph/src/command/compose.rs
```diff
@@ -28,6 +31,10 @@ impl Compose {
 
     fn do_compose(&self) -> BuildResult {
         let supergraph_config = SupergraphConfig::new_from_yaml_file(&self.config_file)?;
+        let federation_version = supergraph_config.get_federation_version();
+        if !matches!(federation_version, FederationVersion::FedOne) {
+            return Err(ConfigError::InvalidConfiguration {message: format!("It looks like '{}' resolved to 'federation_version: {}', whyich doesn't match the current supergraph binary.", &self.config_file, federation_version )}.into());
```
benweatherman, line 36: **nitpick** (same in the fed2 version)

```suggestion
            return Err(ConfigError::InvalidConfiguration {message: format!("It looks like '{}' resolved to 'federation_version: {}', which doesn't match the current supergraph binary.", &self.config_file, federation_version )}.into());
```

install.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
benweatherman, line 1: **nitpick** Would it make sense to throw this in a `./scripts` directory? Might not be worth it if it's short lived but could help declutter the root.

install.sh
```diff
@@ -0,0 +1,17 @@
...
+DEST="$HOME/.rover/bin/supergraph-v$($SRC --version | sed 's/supergraph //g')"
+echo "moving $SRC to $DEST"
+rm -f $DEST
+cp $SRC $DEST
+
+SRC="./federation-2/target/debug/supergraph"
+DEST="$HOME/.rover/bin/supergraph-v$($SRC --version | sed 's/supergraph //g')"
+echo "moving $SRC to $DEST"
+rm -f $DEST
+cp $SRC $DEST
```
benweatherman, line 17: **nitpick** newline

benweatherman (approved):
🐐

benweatherman (approved):

EverlastingBugstopper:
Need to add a test for what happens when a config file is provided with an extra key value

PR #100, opened 2022-04-11
release: composition@v2.0.0

benweatherman:


o0Ignition0o (commented):
Not really sure what is happening [here](https://github.com/apollographql/federation-rs/blob/1dcb3e914129c41ef5ca493c8ce84f32ae20f1cb/apollo-federation-types/src/config/version.rs#L84) but It might be worth a check.
Otherwise lgtm! 🙌

benweatherman:
Thanks for double-checking that @o0Ignition0o! That code should be safe to keep around, since it's checking for a minimum version where we support pinning the composition version. It can probably be removed at some point since it would only be helpful for people trying out the 2.x preview releases (which are mostly internal users).

o0Ignition0o (approved):

PR #104, opened 2022-04-14
fix: QueryPlanner Request race

o0Ignition0o:
fix: QueryPlanner race

This commit fixes a Race that would happen when racing queries against the planner.

It bumps dependencies and introduces a Channel multiplexer. Long story short, Each Request made against deno gets assigned an Uuid, which allows us to send the response to the right requester.

During this refactoring, we've also seen an issue that would happen when the TS runtime would try to return javascript Errors.

Since an Error is not iterable, it would not pass the Deno -> Rust fence because we rely on serialization. We thus catch errros and turn them into an { message, name, stack } structure, that we send to the deno side.

Moreover this commit reworks the Query Planning instanciation and allows us to return clearer error messages.

### DIFF
--- a/federation-2/router-bridge/js-src/plan_worker.ts
+++ b/federation-2/router-bridge/js-src/plan_worker.ts
@@ -62,7 +62,7 @@ const updateQueryPlanner = (schema: string): WorkerResult => {
     const errors = errorArray.map((err) => {
       // We can't import GraphqlError,
       // which would have been less hacky
-      if (!!err.extensions) {
+      if (err?.extensions !== null && err?.extensions !== undefined) {
         return err;
       }
 
--- a/federation-2/router-bridge/js-src/plan_worker.ts
+++ b/federation-2/router-bridge/js-src/plan_worker.ts
@@ -23,6 +23,15 @@ interface Exit {
   kind: PlannerEventKind.Exit;
 }
 type PlannerEvent = UpdateSchemaEvent | PlanEvent | Exit;
+type PlannerEventWithId = {
+  id: string;
+  payload: PlannerEvent;
+};
+
+type WorkerResultWithId = {
+  id?: string;
+  payload: WorkerResult;
+};
 type WorkerResult =
   // Plan result
   ExecutionResult<QueryPlan> | FatalError;
@@ -31,9 +40,9 @@ type FatalError = {
   errors: Error[];
 };
 
-const send = async (payload: WorkerResult): Promise<void> =>
+const send = async (payload: WorkerResultWithId): Promise<void> =>
   await Deno.core.opAsync("send", payload);
-const receive = async (): Promise<PlannerEvent> =>
+const receive = async (): Promise<PlannerEventWithId> =>
   await Deno.core.opAsync("receive");
 
 let planner: BridgeQueryPlanner;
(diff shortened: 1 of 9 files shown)


federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -43,34 +52,59 @@ const updateQueryPlanner = (schema: string): WorkerResult => {
...
+    // The error that has been thrown needs to be sent back
+    // to the rust runtime. In order to do so, it will be serialized.
+    // The code below will allow us to build an object that is JSON serializable,
+    // yet contains all of the information we need
+
+    const errorArray = Array.isArray(err) ? err : [err];
+    const errors = errorArray.map((err) => {
+      // We can't import GraphqlError,
+      // which would have been less hacky
+      if (!!err.extensions) {
```
Geal, line 65: `!!`?

federation-2/router-bridge/src/worker.rs
```diff
@@ -67,26 +87,73 @@ where
...
     }
 
-    pub(crate) async fn send(&self, command: Request) -> Result<(), Error> {
+    pub(crate) async fn send<Request>(&self, request: Request) -> Result<Uuid, Error>
+    where
+        Request: Serialize + Send + Debug + 'static,
+    {
+        let id = Uuid::new_v4();
+
+        let (sender, receiver) = mpsc::channel(1);
```
Geal, line 118: a oneshot channel should be enough here

federation-2/router-bridge/js-src/plan_worker.ts
o0Ignition0o: 🙈

federation-2/router-bridge/src/worker.rs
```diff
@@ -1,39 +1,59 @@
...
-{
-    sender: Sender<Request>,
+#[derive(Serialize, Deserialize, Debug)]
+struct JsonPayload {
+    id: Uuid,
+    payload: serde_json::Value,
+}
+
+pub(crate) struct JsWorker {
+    response_senders: Arc<Mutex<HashMap<Uuid, mpsc::Sender<serde_json::Value>>>>,
```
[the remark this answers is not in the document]
o0Ignition0o: nice catch!

federation-2/router-bridge/src/worker.rs
o0Ignition0o: Oh indeed, it's a leftover from my dashmap tryouts!

federation-2/router-bridge/js-src/plan_worker.ts
o0Ignition0o: fixed in f47bac8

federation-2/router-bridge/src/worker.rs
o0Ignition0o: fixed in f47bac8

federation-2/router-bridge/src/worker.rs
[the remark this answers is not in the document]
o0Ignition0o: fixed in f47bac8

Geal (approved):

federation-2/router-bridge/src/planner.rs
```diff
@@ -148,24 +129,44 @@ impl BridgeError {
...
+}
+
+/// We couldn't instantiate a query planner with the given schema
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct PlannerSetupError {
+    /// The error message
+    pub message: Option<String>,
+    /// The error kind
+    pub name: Option<String>,
+    /// A stqcktrace if applicable
```
bnjjj, line 146:
```suggestion
    /// A stacktrace if applicable
```

federation-2/router-bridge/src/worker.rs
```diff
@@ -1,39 +1,59 @@
...
+        let (response_sender, receiver) = bounded::<JsonPayload>(10_000);
+        let (sender, request_receiver) = bounded::<JsonPayload>(10_000);
+
+        tokio::spawn(async move {
+            while let Ok(json_payload) = receiver.recv().await {
+                let sender = cloned_senders
+                    .lock()
+                    .await
+                    .remove(&json_payload.id)
+                    .expect("TODO");
```
bnjjj, line 46: Better error message ?

bnjjj (approved):

federation-2/router-bridge/src/planner.rs
o0Ignition0o: fixed in 32b1c2e

federation-2/router-bridge/src/worker.rs
o0Ignition0o: fixed in 32b1c2e

PR #92, opened 2022-04-05
wip: usageReportingSignature

o0Ignition0o:

PR #106, opened 2022-04-22
feat: Operation derived data

o0Ignition0o:
Fixes https://github.com/apollographql/router/issues/424

This PR introduces usage reporting to the query planner.


o0Ignition0o:
Here's a rough list of tests I need to add before this is ready to review:

Logging:
- [ ] trace
- [ ] debug
- [ ] info
- [ ] warn
- [ ] error

Usage reporting works:
- [ ] a query and a provided operation name
- [ ] a named query but no provided operation name
- [ ] an anonymous query

Usage reporting works even though query planning failed:
- [ ] parse failure
- [ ] validation failure
- [ ] unknown operation

Feel free to add more tests that seem relevant!

o0Ignition0o:
https://github.com/apollographql/federation-rs/pull/92#discussion_r855713754 shall be adressed as well

o0Ignition0o:
@glasser I gave [your comment](https://github.com/apollographql/federation-rs/pull/92#discussion_r855713754) a bit more thought, and [operationFromDocument](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L1340) can throw several errors, all of them seem to be validation errors, but the unknown operation name one seems to be a bit more specific.

The issue I have is that all errors are of the same type (`Errors`) so I m not quite sure how to do a more futureproof graphql errors <=> usage reporting signature mapping.

I believe this would be a nice followup though

bnjjj (approved):

federation-2/router-bridge/js-src/plan.ts
```diff
@@ -21,63 +47,463 @@ export class BridgeQueryPlanner {
...
-    const operation = operationFromDocument(
-      composedSchema,
-      operationDocument,
-      operationName
+// ---------------------
+
+// Adapted from here
+// https://github.com/apollographql/apollo-server/blob/06dd1171562564f0ca180e74a6fa5530e0389076/packages/apollo-server-core/src/plugin/usageReporting/referencedFields.ts
+
+export function calculateReferencedFieldsByType({
```
Geal, line 169: shouldn't this be in query-planner-js instead, so it can be shared with the gateway?

o0Ignition0o: Absolutely! this will be [refactored eventually](https://github.com/apollographql/federation-rs/pull/92#discussion_r855714172)

federation-2/router-bridge/src/planner.rs
```diff
@@ -147,30 +148,101 @@ pub struct PlannerSetupError {
...
+    /// whether the field is an interface
+    pub is_interface: Option<bool>,
+}
+
+#[derive(Deserialize, Serialize, Debug, PartialEq, Clone)]
+#[serde(rename_all = "camelCase")]
+/// UsageReporting fields, that will be used
+/// to send stats to uplink/studio
+pub struct UsageReporting {
+    /// this is a query related hash that will allow uplink
```
BrynCooke, line 167: Can we put some more detail in the comment.
```
The `stats_report_key` is a unique identifier derived from schema and query.
Metric data is sent to Studio via must be aggregated via grouped key of (`client_name`, `client_version`, `stats_report_key`).

```

BrynCooke (approved):
Looks great. One suggection where the comment could be improved.

federation-2/router-bridge/js-src/plan.ts
```diff
@@ -21,63 +47,463 @@ export class BridgeQueryPlanner {
...
+// In a GraphQL AST (which notably does not contain comments), the only place
+// where meaningful whitespace (or double quotes) can exist is in StringNodes.
+// So to print with reduced whitespace, we:
+// - temporarily sanitize strings by replacing their contents with hex
+// - use the default GraphQL printer
+// - minimize the whitespace with a simple regexp replacement
+// - convert strings back to their actual value We normalize all strings to
+//   non-block strings for simplicity.
+// (Unlike stripIgnoredCharacters, this does not remove commas.)
+function printWithReducedWhitespace(ast: DocumentNode): string {
```
Geal, line 351: this will be fine for now, but maybe let's put a reminder to find a cleaner way sometime :smile:

federation-2/router-bridge/js-src/plan.ts
```diff
@@ -21,63 +47,463 @@ export class BridgeQueryPlanner {
...
+  return minimizedButStillHex.replace(/"([a-f0-9]+)"/g, (_, hex) =>
+    JSON.stringify(Buffer.from(hex, "hex").toString("utf8"))
+  );
+}
+
+// sortAST sorts most multi-child nodes alphabetically. Using this as part of
+// your signature calculation function may make it easier to tell the difference
+// between queries that are similar to each other, and if for some reason your
+// GraphQL client generates query strings with elements in nondeterministic
+// order, it can make sure the queries are treated as identical.
```
Geal, line 375: field order is supposed to be significant, should we really have queries with different order reported in the same way? If this is only used for field level reporting that should be fine, but otherwise the queries should be separated, no?

federation-2/router-bridge/js-src/plan.ts
```diff
@@ -21,63 +47,463 @@ export class BridgeQueryPlanner {
...
+      };
+    },
+    SelectionSet(node: SelectionSetNode): SelectionSetNode {
+      return {
+        ...node,
+        // Define an ordering for field names in a SelectionSet.  Field first,
+        // then FragmentSpread, then InlineFragment.  By a lovely coincidence,
+        // the order we want them to appear in is alphabetical by node.kind.
+        // Use sortBy instead of sorted because 'selections' is not optional.
+        selections: sortBy(node.selections, "kind", "name.value"),
```
Geal, line 404: should the alias be included in sorting keys?

federation-2/router-bridge/js-src/plan.ts
```diff
@@ -21,63 +47,463 @@ export class BridgeQueryPlanner {
...
+    return sortBy(items, ...iteratees);
+  }
+  return undefined;
+}
+
+// removeAliases gets rid of GraphQL aliases, a feature by which you can tell a
+// server to return a field's data under a different name from the field
+// name. Maybe this is useful if somebody somewhere inserts random aliases into
+// their queries.
+function removeAliases(ast: DocumentNode): DocumentNode {
```
Geal, line 451: Aliases are also used if we want to return the same field multiple times but with different directives, should they be removed in that case?

federation-2/router-bridge/js-src/plan.ts
o0Ignition0o: Yes, I ve documented where the code comes from so we know where to refactor it. this code is used in the gateway and in the bridge so we can be sure the usageReportingSignature stays the same when users migrate.

federation-2/router-bridge/js-src/plan.ts
o0Ignition0o: I think you're correct, however the code has been copy pasted from the apollo-server implementation in order to make sure the bridge reports usage in the same way as the server

federation-2/router-bridge/js-src/plan.ts
o0Ignition0o: good question, possibly worth a discussion with the federation team ?

federation-2/router-bridge/js-src/plan.ts
o0Ignition0o: This code has been duplicated from the gateways to make sure the behavior remains the same. I agree it's however worth discussing once we refactor (or riir?) it

federation-2/router-bridge/src/planner.rs
```diff
@@ -147,30 +148,101 @@ pub struct PlannerSetupError {
...
+    /// Usage reporting related data such as the
+    /// operation signature and referenced fields
+    pub usage_reporting: Option<UsageReporting>,
+}
+
+/// The payload if the plan_worker invocation failed
+#[derive(Debug, Clone)]
+pub struct PlanErrors {
+    /// The errors the plan_worker invocation failed with
+    pub errors: Arc<Vec<PlanError>>,
```
Geal, line 201: why is there an Arc here?

Geal (approved):

federation-2/router-bridge/js-src/plan.ts
Geal: alright then

federation-2/router-bridge/src/planner.rs
o0Ignition0o: It's because routerservice boxerrors need Clone, i totally agree this probably doesnt belong here tho, shall i change it ?

glasser:
> @glasser I gave [your comment](https://github.com/apollographql/federation-rs/pull/92#discussion_r855713754) a bit more thought, and [operationFromDocument](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L1340) can throw several errors, all of them seem to be validation errors, but the unknown operation name one seems to be a bit more specific.
> 
> The issue I have is that all errors are of the same type (`Errors`) so I m not quite sure how to do a more futureproof graphql errors <=> usage reporting signature mapping.
> 
> I believe this would be a nice followup though

Hi @o0Ignition0o ! Congrats on getting this merged. Sorry, I got a few days behind in my email/code reviews and didn't get to this yet.

Here's the way I think about the three special error types.

Whether or not something is a GraphQLParseFailure is a pure function of the operation itself. It does not depend on the schema or on the operation name string optionally supplied alongside the operation in the request.

Whether or not something is a GraphQLValidationFailure is a function of the operation and the schema.  It's about "this legally-parsing bit of GraphQL syntax doesn't actually work with this schema".  But it is *not* a function of the operation name string optionally supplied alongside the operation in the request.

Finally, whether or not something is a GraphQLUnknownOperationName is a function of the operation and the schema *and the separately supplied operation name*. So there are at least two distinct ways that this can fail.  One of them is that you provide an operation name but that operation name doesn't match an operation name in the document. And the other is that the document has multiple operations but you don't provide an operation name to select between them.

(none of these cases cover "the variables you provided alongside the operation don't match the expected ones" btw; that is treated as just a part of execution failing even though it's more of a client-side issue.)

In your code as written, you're only recognizing the first case as GraphQLUnknownOperationName. But you need to recognize the second case too.  That one is the one where operationFromDocument throws an error like `'Must provide operation name if query contains multiple operations.'`.

My suggestion was that probably any operationFromDocument throw represents GraphQLUnknownOperationName since the other paths that throw *probably* are actually conditions that would have already been caught by graphql-js `validate`.  It's fair to disagree with that since there are a ton of possible throws in operationFromDocument :) but at the very least, the "must provide operation name" does need to be caught.

(The other case to consider is the `No operation found in provided document.` case, which happens if you don't provide an operation name *and* there's no operation at all in the document.  Skimming [the graphql-js validation rules](https://github.com/graphql/graphql-js/blob/ea1c34720d1b5cc8691856e392aae1aa453339e1/src/validation/specifiedRules.ts#L77-L102) I don't think there's an explicit rule for "there must be at least one operation". However, the graphql-js *parser* requires there to be at least one top-level definition (see [parseDocument](https://github.com/graphql/graphql-js/blob/ea1c34720d1b5cc8691856e392aae1aa453339e1/src/language/parser.ts#L206-L215) and note that [this.many](https://github.com/graphql/graphql-js/blob/ea1c34720d1b5cc8691856e392aae1aa453339e1/src/language/parser.ts#L1519-L1524) matches a non-empty list of things. And the validation rules do include [a rule that requires all definitions to be executable (not type-system)](https://github.com/graphql/graphql-js/blob/main/src/validation/rules/ExecutableDefinitionsRule.ts) which is to say operations and fragments.  So the only way to have no operations is to have a document consisting only of fragments. But there's [another rules](https://github.com/graphql/graphql-js/blob/main/src/validation/rules/NoUnusedFragmentsRule.ts) that says all fragments must be used (perhaps transitively) in an operation. Thus, a document without any operations is either a parse failure or a validation failure and so the `No operation found in provided document.` case should not be possible. I encourage you to double check my logic here!)

So my request is that you ensure that the `'Must provide operation name if query contains multiple operations.'` case results in `GraphQLUnknownOperationName` too.  This is something like `query X { f } query Y { f }` with no operationName provided. I also would encourage you to add comments in operationFromDocument making it clear that the usage reporting code depends on the particular error messages thrown (or to use some other mechanism of detecting these two errors).

Issue #80, opened 2022-03-17
add `composition-latest-0` and `composition-latest-2` tags when running `cargo xtask tag`


EverlastingBugstopper:
in orbiter, we need a way to get the latest version of the `supergraph` CLIs for either v0 or v2. easiest way to do this is with a tag that we update for each. i've manually created these tags based on what's here but we need to automate updating it. i've already created the first versions of these tags locally. until this is closed we should manually update them to point to the correct place.

EverlastingBugstopper:
thinking on how this would work in practice - we don't want to update the tags until the tarball and installer endpoints for a new version are live. therefore, i think we should have something maybe like `cargo xtask promote supergraph@v{version}` run _after_ the packages are uploaded to GitHub, which would run the following in [`GitRunner`](https://github.com/apollographql/federation-rs/blob/main/xtask/src/tools/git.rs): 

```
git tag -d composition-latest-2 && git tag -a composition-latest-2 -m v{version} && git push --tags -f
```

benweatherman:
The `cargo xtask promote` flow makes sense, and gives us a chance to test before making it live. I think the jankiness is great for now 😈

EverlastingBugstopper:
this has been superseded by #197 - we should use renovate in the rover repo to handle updating these

PR #126, opened 2022-06-01
Update instructions for releasing supergraph component

benweatherman:
1. Use `--debug` in `cargo xtask dist` so `Cargo.lock` can be updated. Without that param, you'll get the following error of sadness:
    ```
    ➜ cargo xtask dist
        Finished dev [unoptimized + debuginfo] target(s) in 0.04s
        Running `target/debug/xtask dist`
    info: running `cargo build --release --locked --target x86_64-apple-darwin` in `/Users/weatherman/code/apollographql/federation-rs/federation-1`
    info: running `cargo build --release --locked --target x86_64-apple-darwin` in `/Users/weatherman/code/apollographql/federation-rs/federation-2`
    info: running `cargo build --release --locked --target x86_64-apple-darwin` in `/Users/weatherman/code/apollographql/federation-rs/federation-2/router-bridge`
    error: the lock file /Users/weatherman/code/apollographql/federation-rs/federation-2/Cargo.lock needs to be updated but --locked was passed to prevent this
    If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.

    Error: Could not run command in `/Users/weatherman/code/apollographql/federation-rs/federation-2/router-bridge`

    Caused by:
        0: Encountered an issue while executing `cargo build --release --locked --target x86_64-apple-darwin`
        1: Exited with status code 101
    ```
1. Add instructions for tagging `latest-2` release so the supergraph component is downloaded by rover

### DIFF
--- a/federation-2/harmonizer/RELEASE_CHECKLIST.md
+++ b/federation-2/harmonizer/RELEASE_CHECKLIST.md
@@ -14,7 +14,7 @@ None of the `federation-rs` packages currently maintain changelogs as they are l
 1. Run `PUBSLUG=composition@v{version}` where `{version}` is the new version you're bumping to. The major version should NOT be 0, it should be 2.
 1. Run `git checkout main && git stash && git pull && git checkout -b $PUBSLUG`.
 1. Update the version of `@apollo/composition` in `./harmonizer-2/package.json`
-1. Run `cargo xtask dist` from the root of `federation-rs`
+1. Run `cargo xtask dist --debug` from the root of `federation-rs`
 1. Push up a commit containing the version bumps with the message `release: $PUBSLUG`
 1. Wait for tests to pass on the PR
 1. Merge your PR to `main`
@@ -23,7 +23,14 @@ None of the `federation-rs` packages currently maintain changelogs as they are l
 
 1. Once merged, run `git checkout main && git pull`
 1. Run `cargo xtask tag --package $PUBSLUG --real-publish`
+   - **NOTE** If you get an error, you might need to run `git fetch --tags -f` to ensure the remote tags match the local tags
 1. Wait for CI to build and publish `harmonizer` to crates.io and `supergraph` to `federation-rs` GitHub releases.
+
+### Tag the `latest-2` release so rover automatically downloads the new version
+
+1. Run `git tag -d composition-latest-2 && git tag -a composition-latest-2 -m v{version} && git push --tags -f` replacing `{version}` with the version of composition.
+
+TODO: tag latest-2 / latest-1
 
 ## Troubleshooting a release
 
--- a/federation-2/harmonizer/RELEASE_CHECKLIST.md
+++ b/federation-2/harmonizer/RELEASE_CHECKLIST.md
@@ -23,7 +23,6 @@ None of the `federation-rs` packages currently maintain changelogs as they are l
 
 1. Once merged, run `git checkout main && git pull`
 1. Run `cargo xtask tag --package $PUBSLUG --real-publish`
-   - **NOTE** If you get an error, you might need to run `git fetch --tags -f` to ensure the remote tags match the local tags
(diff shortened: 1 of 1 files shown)


federation-2/harmonizer/RELEASE_CHECKLIST.md
```diff
@@ -23,8 +23,13 @@ None of the `federation-rs` packages currently maintain changelogs as they are l
 
 1. Once merged, run `git checkout main && git pull`
 1. Run `cargo xtask tag --package $PUBSLUG --real-publish`
+   - **NOTE** If you get an error, you might need to run `git fetch --tags -f` to ensure the remote tags match the local tags
```
EverlastingBugstopper, line 26: this _should_ happen automatically in xtask tag ([src](https://github.com/apollographql/federation-rs/blob/main/xtask/src/tools/git.rs#L94))- did you run into this issue? i am of the belief (perhaps a mistaken belief) that if you run `git fetch --tags -f` it will overwrite any local tags with the tags on the remote, but it will not delete any local tags that you have on your machine. So, if you had a local tag that you didn't want to push, `git fetch --tags -f` would leave you in a state where the next time you `git push --tags`, it would push any local tag you created. (in the past i've run into situations where tags end up on other dev machines and just, always inevitably get pushed even if you deleted them in the past. they are very sticky, ick... thus, this little ritual) 

so in `xtask tag` we actually do the following (there may be a better way):

1) list all the tags on the local machine
2) iterate through those and delete every single one of them until there are none left, kill them with fire etc. etc.
3) fetch the tags from the remote
4) create another tag locally
5) push the tag, ensuring it is only one tag that is being pushed at a time

federation-2/harmonizer/RELEASE_CHECKLIST.md
```diff
@@ -23,8 +23,13 @@ None of the `federation-rs` packages currently maintain changelogs as they are l
...
 1. Wait for CI to build and publish `harmonizer` to crates.io and `supergraph` to `federation-rs` GitHub releases.
 
+### Tag the `latest-2` release so rover automatically downloads the new version
+
+1. Run `git tag -d composition-latest-2 && git tag -a composition-latest-2 -m v{version} && git push --tags -f` replacing `{version}` with the version of composition.
```
EverlastingBugstopper, line 31: this is great to document - I think it only existed in slack DMs before. thank you for doing so. if you wanna take a crack at #80 we could make this a part of `xtask tag`? Can do in a follow-up PR for sure, I'd be happy to help out w/reviews and/or pairing if you'd like!

federation-2/harmonizer/RELEASE_CHECKLIST.md
```diff
@@ -14,7 +14,7 @@ None of the `federation-rs` packages currently maintain changelogs as they are l
 1. Run `PUBSLUG=composition@v{version}` where `{version}` is the new version you're bumping to. The major version should NOT be 0, it should be 2.
 1. Run `git checkout main && git stash && git pull && git checkout -b $PUBSLUG`.
 1. Update the version of `@apollo/composition` in `./harmonizer-2/package.json`
-1. Run `cargo xtask dist` from the root of `federation-rs`
+1. Run `cargo xtask dist --debug` from the root of `federation-rs`
```
EverlastingBugstopper, line 17: 👍🏼 👍🏼

EverlastingBugstopper (approved):
🎉 lgtm! left a few comments, none of which are blocking

federation-2/harmonizer/RELEASE_CHECKLIST.md
EverlastingBugstopper: i left [a comment](https://github.com/apollographql/federation-rs/issues/80#issuecomment-1144225559) on the approach i think we would want

benweatherman: Sweet! I'd be down to take a stab at that, though in a different PR.

PR #139, opened 2022-06-10
chore: attempt to force host_cpu to arm

EverlastingBugstopper:
by setting `GN_ARGS="host_cpu=\"arm64\"` and hoping for the best (from [this issue](https://github.com/denoland/rusty_v8/issues/520)). i'm hoping we don't need to build v8 from source and that this will just correctly create the snapshots

### DIFF
--- a/xtask/src/target.rs
+++ b/xtask/src/target.rs
@@ -48,6 +48,10 @@ impl Target {
         Self::MacOSAmd64 == *self || Self::MacOSAarch64 == *self
     }
 
+    pub(crate) fn is_arm_macos(&self) -> bool {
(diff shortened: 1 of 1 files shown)


EverlastingBugstopper:
tested this out w/o running in CI by cross-compiling on my intel macbook, and using `scp` to copy the binary to a remote M1. still getting the "unreachable" C error message unfortunately.

EverlastingBugstopper:
this will not work because there is no way to get v8 to generate ARM bytecode for its snapshots unless it's running natively on an ARM device - we will only be providing x86_64 plugins because you can't run ARM mac machines in CI

PR #145, opened 2022-07-08
`@defer` implementation

Geal:
following updates in https://github.com/apollographql/federation/pull/1958


Geal:
this should be ok to review now, but maybe there should be a separate PR to update federation dependencies everywhere to 2.1.0-alpha3 as was done in https://github.com/apollographql/federation-rs/pull/149

o0Ignition0o (approved):

PR #150, opened 2022-08-05
use npm packages i published to s3

abernix:
using s3 for this because codesandbox just randomly deleted the
`federation-internals` package from their registry.

i assure you this is only temporary.


o0Ignition0o:
Merging this as agreed with @abernix and @Geal since it will be a temporary fix for https://github.com/apollographql/router/issues/1467

PR #152, opened 2022-08-12
release: composition@v2.1.0-alpha.4

clenfest:


benweatherman (approved):
🐐

apollo-federation-types/src/build/error.rs
```diff
@@ -5,7 +5,7 @@ use std::{
 
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
```
o0Ignition0o, line 8: oh that would be the new clippy lint :D

o0Ignition0o (approved):
LGTM!

I can't find any sensible reason why we derive PartialEq and Eq on all these types, except maybe for tests... It might be worth investigating, especially now that we're doing snapshot tests that work on json anyway.

PR #169, opened 2022-09-05
router-bridge: better error reporting for unexpected errors

garypen:
This doesn't address the root cause of an unexpected error, but it makes
the transtion of the error back through the router a little cleaner.


garypen:
It doesn't "fix" https://github.com/apollographql/router/issues/1519, but it makes the error reporting cleaner.

federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -191,10 +191,17 @@ async function run() {
         }
       } catch (e) {
         logger.warn(`an error happened in the worker runtime ${e}\n`);
+        let unexpectedError = {
+          name: "unexpectedError",
+          message: `${e}`,
+          extensions: {
+            code: "an error happened in the worker runtime",
```
abernix, line 198: Does this message end up being client facing?  If so, I don't think this will be meaningful enough as it is right now.

Additionally, `code` should be a machine readable code — it's typically something like `GRAPHQL_PARSE_FAILED`.   See the list [of some known codes here](https://www.apollographql.com/docs/apollo-server/data/errors/#error-codes), though I think we have some in Router, too.  Point being, I don't think it should be a message string, which would be on another property (perhaps `message`?)

federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -204,9 +211,16 @@ async function run() {
       }
     } catch (e) {
       logger.warn(`plan_worker: an unknown error occured ${e}\n`);
+      let unexpectedError = {
+        name: "unexpectedError",
```
abernix, line 215: It will be unexpected to us, but it potentially has a [meaningful `name` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name) on the `e` (Error).  More on this in my next comment

federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -204,9 +211,16 @@ async function run() {
...
+        message: `${e}`,
```
abernix, line 216: This notation (``` `${e}` ```) will result in invoking `e.toString()` on the `e` object _if that method is defined_.  In **most** cases, that is defined as `self.message + "\n" + self.stack`, where [`stack` is the JS part of the stack](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack).

It's worth noting that the stack will be largely bundled symbols and relatively hard to interpret meaning and map it to code.  I expect it will be most valuable to us, rather than the client, whereas `message` might have meaning.

**Perhaps** to provide a nicer message, we might want to capture [`e.message` (in particular)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message) here and have another property on the error object which captures the string serialized message that includes `stack`?  Perhaps even a third dedicated property for `stack`?

federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -191,10 +191,17 @@ async function run() {
         }
       } catch (e) {
         logger.warn(`an error happened in the worker runtime ${e}\n`);
+        let unexpectedError = {
+          name: "unexpectedError",
+          message: `${e}`,
+          extensions: {
+            code: "an error happened in the worker runtime",
+          },
```
abernix, line 199: Building on the other quasi-standard properties of the GraphQL error ([see the specification](https://spec.graphql.org/draft/#sec-Errors.Error-result-format) here) maybe we want something like this:

```json5
"extensions": {
  "code": "QUERY_PLANNING_FAILED",
  "exception": {
    "stacktrace": [ /* I think this can be ...(err.toString().split(/\n)) ? */ ]
  }  
}
```

abernix (changes requested):

abernix:
I suspect this will be a great improvement, btw.  I'm also curious if we should just log the details to the server log but only expose a portion of the details to the client.

garypen:
I've tried to incorporate some of the feedback, but I don't write typescript so the quality may/will be low.

federation-2/router-bridge/js-src/plan_worker.ts
garypen: I tried doing this. No idea if the syntax is correct.

federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -191,10 +191,21 @@ async function run() {
...
       } catch (e) {
         logger.warn(`an error happened in the worker runtime ${e}\n`);
+
+        const unexpectedError = {
+          name: e.name || "unknown",
+          message: e.message || "",
+          extensions: {
+            code: "QUERY_PLANNING_FAILED",
+            exception: {
+              stacktrace: [e.toString().split(/\n/)],
```
abernix, line 201: I know this is still work in progress, but note that this isn't right and will result in an array within an array.  Removing the outside `[`/`]` would be correct.

abernix, line 201: Thoughts on just creating an actual `GraphQLError` here (which accepts I think everything besides `name` already) rather than creating a object that is roughly shaped like the `GraphQLError`?

I'm curious what wouldn't work about that?

garypen (commented):
approve.

federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -204,9 +216,21 @@ async function run() {
       }
     } catch (e) {
       logger.warn(`plan_worker: an unknown error occured ${e}\n`);
```
abernix, line 218:
```suggestion
      logger.warn(`plan_worker: an unknown error occurred ${e}\n`);
```

abernix (approved):

abernix:
In the end, I'd say this is more than just slightly improved!  Thanks!

PR #174, opened 2022-09-08
try out an idea for protecting the query planner from abuse

garypen:
The idea is to maintain a HashSet of queries where we couldn't notify our requester that query planning has completed.

They become, effectively, blacklisted and the next time we receive that request, we just reject it.

This would help a lot with complex query planning.

The are some drawbacks with the approach though. There are many legitimate resons for a requester to disappear, so it seems harsh to always blacklist (see comments in code for ways to mitigate).

Anyway, I think it's worth exploring further and as a side effect I got rid of Uuid and rely on the rust hasher for ID generation.

### DIFF
--- a/federation-2/Cargo.lock
+++ b/federation-2/Cargo.lock
@@ -849,7 +849,6 @@ dependencies = [
  "tower-service",
  "tracing",
  "tracing-test",
- "uuid",
  "which",
 ]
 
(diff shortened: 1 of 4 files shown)


garypen:
Possibly a better idea.
Store query plans that we fail to deliver with their id in a HashMap. Next time they are requested, remove the previously undelivered result from the map and return this so it can be cached as a successful result in the router...

garypen:
#176 replace this.

bnjjj (approved):
I think it would be worth to have a follow up PR to implement a size limited cache. But for now that's good enough !

PR #175, opened 2022-09-09
Try to use the GraphQLError class to produce the error

abernix:
Just to test this approach


federation-2/router-bridge/js-src/plan_worker.ts
```diff
@@ -215,18 +214,17 @@ async function run() {
         });
       }
     } catch (e) {
-      logger.warn(`plan_worker: an unknown error occured ${e}\n`);
+      logger.warn(`plan_worker: an unknown error occurred ${e}\n`);
```
abernix, line 217: Also fix a typo.

bnjjj:
I already tried, it produces a bad bundle.js with require inside and deno is unable to launch with `require`

abernix:
> I already tried, it produces a bad bundle.js with require inside and deno is unable to launch with `require`

Yeah, I know.  I wanted to see why.  We already use `GraphQLError` in other files.

abernix:
I am concerned this doesn't work and I'm going to leave this PR open for now.

PR #184, opened 2022-09-14
release router-bridge@v2.1.2-alpha.2

jeffjakub:


abernix (changes requested):
You'll need to also bump the version of the packages in the `router-bridge/package.json` (and corresponding `package-lock.json` - `npm i --package-lock-only` will suffice!), in the same way as `harmonizer/package*.json`.  There are two relevant `@apollo/` 2.1.2 packages in there:

https://github.com/apollographql/federation-rs/blob/b000b6091ae0dc88a60cada45da99446aed5350b/federation-2/router-bridge/package.json#L31-L32

abernix:
Roughly, I'd expect something like [this](https://github.com/apollographql/federation-rs/commit/beb932bedc1ddf774de9909bc8739fa693abb914) but with this `router-bridge` package.  I could imagine that there is some documentation that needs to be updated if this isn't written down.

abernix:
Since we're trying to release today, I'm going to go ahead and push this change up and finish this PR.

abernix:
Ah, I just caught this but I guess the title of this PR / merge commit would have been `release: router-bridge@v0.1.7` but all good.

abernix:
We _should_ actually just align that with the underlying federation query planner version in the next release though, probably.

jeffjakub:
Good call out, I'll make some notes and submit a PR. Router-bridge steps differ a bit from harmonizer. I'll submit a separate PR with updates for the router-bridge release checklist.

PR #185, opened 2022-09-15
router-bridge: Disable Deno snapshotting when docs.rs builds docs

SimonSapin:
This attempts to fix build errors like https://docs.rs/crate/apollo-router/1.0.0-rc.0/builds/629200 by imitating what Deno does in https://github.com/denoland/deno/blob/6bb72a80863ac3913d32ea21aae32dd327ce6b71/runtime/build.rs#L228-238

### DIFF
--- a/federation-2/router-bridge/Cargo.toml
+++ b/federation-2/router-bridge/Cargo.toml
@@ -45,3 +45,11 @@ tracing-test = "0.2.1"
 [build-dependencies]
 deno_core = "0.142.0"
 which = "4.2.2"
+
+[features]
+# "fake" feature to disable V8 usage when building on docs.rs
+# See ./build.rs
+docsrs = []
+
+[package.metadata.docs.rs]
+features = ["docsrs"]
--- a/federation-2/router-bridge/Cargo.toml
+++ b/federation-2/router-bridge/Cargo.toml
@@ -49,7 +49,7 @@ which = "4.2.2"
 [features]
 # "fake" feature to disable V8 usage when building on docs.rs
 # See ./build.rs
-docsrs = []
+docs_rs = []
 
 [package.metadata.docs.rs]
 features = ["docsrs"]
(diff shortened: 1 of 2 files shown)


SimonSapin:
I’ve tested that `cargo doc --features docsrs` succeeds on my laptop, but we’ll have to wait until `router-bridge` is next published to crates.io to see what happens on actual docs.rs build servers.

federation-2/router-bridge/Cargo.toml
```diff
@@ -45,3 +45,11 @@ tracing-test = "0.2.1"
 [build-dependencies]
 deno_core = "0.142.0"
 which = "4.2.2"
+
+[features]
+# "fake" feature to disable V8 usage when building on docs.rs
+# See ./build.rs
+docsrs = []
```
garypen, line 52: NIT: Improved readability
```suggestion
docs_rs = []
```

federation-2/router-bridge/build.rs
```diff
@@ -69,7 +66,20 @@ fn update_bridge(current_dir: &Path) {
         .success());
 }
 
+#[cfg(feature = "docsrs")]
 fn create_snapshot(out_dir: &Path) {
+    // If we're building on docs.rs we just create
+    // and empty snapshot file and return, because `rusty_v8`
```
garypen, line 72:
```suggestion
    // an empty snapshot file and return, because `rusty_v8`
```

Geal (approved):

garypen (commented):
Looks good.

federation-2/router-bridge/Cargo.toml
```diff
@@ -45,3 +45,11 @@ tracing-test = "0.2.1"
 [build-dependencies]
 deno_core = "0.142.0"
```
Geal, line 46: should deno_core be marked as optional?

o0Ignition0o (approved):

federation-2/router-bridge/Cargo.toml
SimonSapin: No because `cargo doc` runs the equivalent of `cargo check`. It’s only the C++ bits of V8 that are disabled apparently, all the Rust bits of Deno and rusty_v8 and still there.

PR #188, opened 2022-09-21
Updating the harmonizer and router-bridge release documentation 

jeffjakub:
Two items:
1. Update the env variable used in harmonizer and router-bridge to be unique.
2. Update the router-bridge release doc to indicate we need to update the package.json.


abernix (approved):
Great changes! And thanks for opening this as a separate PR/improvement!

PR #198, opened 2022-10-07
add support for deno memory controls in the router-bridge

garypen:
This PR introduces support setting a minimum heap size for deno. An experimental flag:
  APOLLO_ROUTER_BRIDGE_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE
is now supported which allows a default, initial heap limit for deno to be specified. If not specified, a default of 256M is used (which I believe is lower than the usual default of 1024M) and appropriate for most supergraph activities.

If introspection is enabled, the setting also applies to the deno introspection execution environment.

If the deno heap is running low, then it will be expanded by 1.25. There is no upper limit, since we rely on the execution environment to let us know if we run out of memory.

Details about heap activities are logged.

This facility should be helpful for those situations where memory resources are extremely constrained.

### DIFF
--- a/federation-2/router-bridge/src/js.rs
+++ b/federation-2/router-bridge/src/js.rs
@@ -8,6 +8,58 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::sync::mpsc::{channel, Sender};
 
+// A reasonable default starting limit for our deno heap.
+const APOLLO_ROUTER_BRIDGE_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE_DEFAULT: &str = "256";
+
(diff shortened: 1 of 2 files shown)


SimonSapin (approved):

EverlastingBugstopper:
This is great!

As a follow-up - I really wish that we could separate out the runtime bits here into a separate crate so that these improvements/configurations would happen in both `router-bridge` and the fed1/fed2 composition plugins.

garypen:
> This is great!
> 
> As a follow-up - I really wish that we could separate out the runtime bits here into a separate crate so that these improvements/configurations would happen in both `router-bridge` and the fed1/fed2 composition plugins.

Thanks Avery. That seems like a good follow-up suggestion. Once we have some good experience with this change we can look at doing that.

PR #200, opened 2022-10-10
Expose hint levels

### LINKED ISSUE #102: feat(composition): add hint levels to output types
hints don't need to be addressed! they're helpful to have, but having them as warnings may push folks to want to fix/remove them. we should just call them hints instead and let people keep them around.
these warning prefixes can be found [here](../blob/main/src/command/output.rs)

pcmanus:
Composition hints in federation 2 are generated with a "level" (which can currently be one of of "DEBUG", "INFO" or "WARN"). This commit modify harmonizer to include those levels in the output and the type definitions to decode them, and include them in the `supergraph` output for future use by `rover`.

Fixes #102

### DIFF
--- a/apollo-federation-types/src/build/hint.rs
+++ b/apollo-federation-types/src/build/hint.rs
@@ -1,10 +1,37 @@
 use serde::{Deserialize, Serialize};
+use regex::Regex;
+use lazy_static::lazy_static;
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+pub struct BuildHintLevel {
+    /// Value of the hint level. Higher values correspond to more "important" hints.
+    pub value: u16,
+
+    /// Readable name of the hint level
+    pub name: String,
+}
+
+impl BuildHintLevel {
+    pub fn warn() -> Self { Self { value: 60, name: String::from("WARN") } }
+    pub fn info() -> Self { Self { value: 40, name: String::from("INFO") } }
+    pub fn debug() -> Self { Self { value: 20, name: String::from("DEBUG") } }
+}
+
 
 /// BuildHint contains helpful information that pertains to a build
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct BuildHint {
     /// The message of the hint
+    /// This will usually be formatted as "[<hint code>] <message details>" and the
+    /// `BuildHint::extract_code_and_message` method can be used to extract the components of this
+    /// message.
     pub message: String,
+
+    /// The level of the hint
+    // We should always get a level out of recent harmonizer, but older one will not have it and we
+    // default to "INFO".
+    #[serde(default="BuildHintLevel::info")]
+    pub level: BuildHintLevel,
 
     /// Other untyped JSON included in the build hint.
     #[serde(flatten)]
@@ -12,10 +39,36 @@ pub struct BuildHint {
 }
 
 impl BuildHint {
-    pub fn new(message: String) -> Self {
+    pub fn new(message: String, level: BuildHintLevel) -> Self {
         Self {
             message,
+            level,
             other: crate::UncaughtJson::new(),
+        }
+    }
+
+    pub fn debug(message: String) -> Self {
+        Self::new(message, BuildHintLevel::debug())
+    }
+
+    pub fn info(message: String) -> Self {
+        Self::new(message, BuildHintLevel::info())
+    }
+
+    pub fn warn(message: String) -> Self {
+        Self::new(message, BuildHintLevel::warn())
+    }
+
+    /// Extracts the underlying code and "raw" message of the hint.
+    pub fn extract_code_and_message(&self) -> (String, String) {
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"^\[(\w+)\] (.+)").unwrap();
+        }
+        let maybe_captures = RE.captures(&self.message);
+        if let Some(captures) = maybe_captures {
+            (captures.get(1).unwrap().as_str().to_string(), captures.get(2).unwrap().as_str().to_string())
+        } else {
+            (String::from("UNKNOWN"), self.message.clone())
         }
     }
 }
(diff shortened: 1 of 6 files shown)


apollo-federation-types/src/build/hint.rs
```diff
@@ -1,23 +1,76 @@
...
+    }
+
+    /// Extracts the underlying code and "raw" message of the hint.
+    pub fn extract_code_and_message(&self) -> (String, String) {
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"^\[(\w+)\] (.+)").unwrap();
+        }
+        let maybe_captures = RE.captures(&self.message);
+        if let Some(captures) = maybe_captures {
+            (captures.get(1).unwrap().as_str().to_string(), captures.get(2).unwrap().as_str().to_string())
```
EverlastingBugstopper, line 69: is there a more structured way that we can get this data from the underlying JS library? I don't love the need to use a regex here for this - if we could instead put that info in a JSON object or something instead of parsing out the string I think this will be more robust

federation-2/harmonizer/deno/do_compose.js
```diff
@@ -56,7 +56,10 @@ try {
   let hints = [];
   if (composed.hints) {
     composed.hints.map((composed_hint) => {
-      hints.push({ message: composed_hint.toString() });
+      hints.push({
+        message: composed_hint.toString(),
+        level: composed_hint.definition.level,
```
EverlastingBugstopper, line 61: it looks like we are putting that info right into the JSON object here in the JS code - can we just do something like 

```rust
let data = serde_json::from_string(data_from_js);
let message = data["message"];
let level = data["level"];
```

?

apollo-federation-types/src/build/hint.rs
pcmanus: I'm not a huge fan either, but I was worried about either making it weird to use, or have backward compatibility problems.

Let me however remark that this method is not about the hint "level". A hint has fundamentally 3 components at the moment:
1. the `message`, which describe the particular hint.
2. the `code`, which categorize hints (same as we have for errors).
3. the `level`, which is `WARN`, `INFO` or `DEBUG` for now.

The `level`, that is introduced by this patch, is shipped as a separate field in the underlying JSON object, and thus deserialized separately, so no worry on this one.

But what this method is trying to do is give access to the `message` and `code` separately. Because the issue is that unfortunately, instead of having the `message` and `code` as separate fields in the underlying JSON, `do_compose.js` has been using the `Hint.toString()` method to generate a single pre-formatted string of the form `"[<code>] <message>"` and calling this "message".

So first, I acknowledge that this isn't strictly hint level related and so one could make the argument this shouldn't be in this patch. But I think the ultimate goal should be that in the rover UX, errors and hints can be displayed in a consistent way, where errors can just be considered as the highest "hint level" (in fact, we do hope to "merge" hints and errors in the code at some point in the future to clean things up). But doing that cleanly implies rover can access the hint `code` and `message` separately, the same way it does for errors, so it can maybe colorize the `code` and/or format it however it pleases. This could also allow flexibility like grouping messages by `code` for both hints and errors, or filtering certain `code`, etc.

Long story short, the current `BuildHint.message` field of `hint.rs` is currently unfortunately not really just the hint `message`, it is a pre-formatted string with both the `code` and (raw) `message` and this method is meant to provide a way out of this.

In theory, a better alternative would be to change `BuildHint.message` to only be the actual hint `message` and to add a `BuildHint.code` for the code, and to maybe also ensure those are separate in the underlying JSON object so it's easier to do, but I was kind of worried about breaking things by doing those things. So the idea of this method was to keep `BuiltHint.message` unchanged so we don't break existing usages, but to allow future usages to be able to access the `code` and `message` components individually.

Anyway, hope this clarify. If you're not worry about the concerns around breaking existing code, I'm happy to change all this, but I think it would require some care around versioning this properly, because if we change this, currently rover version would need to not use federation-rs versions with such changes (or rather, if they do, the hint will suddenly not be displayed with their code since it wouldn't be part of `BuildHint.message` anymore).

EverlastingBugstopper: I don't think we have to worry about breaking existing code! We should definitely change the underlying library code away from the `.toString` here - the only thing that matters is that we don't _remove_ fields from the JSON that's actually emitted by the `supergraph` binary. Everything else is an internal contract - we only need to hold up the end to end guarantee.

pcmanus: > the only thing that matters is that we don't remove fields from the JSON that's actually emitted by the `supergraph` binary

But currently, the JSON of the `supergraph` binary has the `message` field for hints that is the pre-formatted string containing both the "raw" message and the code. And existing rover versions kind of assume this is what this field contains, but we ultimately want rover to be able to access separately and not as a pre-formatted string. So if we do the "cleanest" change here without backward compatibility concerns, we would be changing that `message` field to now be only the "raw" message, without code, and the code would be a new field. Technically this is not a field removal, but it's changing what the field holds, and I assume that would be an issue too?

The alternative could be keep the `message` field containing the pre-formatted string (we can still stop having the pre-formatted string in the JSON _from javascript_, but the `supergraph` binary itself would be the one formatting the string for its own JSON output), but also add 2 new fields, `code` and `raw_message`, with the `message` field essentially becoming deprecated but kept for backward compatibility.

Does that make sense, or am I overcomplicating this?

EverlastingBugstopper: > Technically this is not a field removal, but it's changing what the field holds, and I assume that would be an issue too?

Yeah - I think you're right here, we want older versions to still be able to work. 

> The alternative could be keep the message field containing the pre-formatted string (we can still stop having the pre-formatted string in the JSON from javascript, but the supergraph binary itself would be the one formatting the string for its own JSON output), but also add 2 new fields, code and raw_message, with the message field essentially becoming deprecated but kept for backward compatibility.

Old versions of Rover will discard any new unused JSON fields (we should test this to verify of course), but this approach sounds perfectly reasonable to me!

pcmanus: Thanks for the feedback. I'll try to update the patch accordingly ... when I find a bit of time for it.

PR #202, opened 2022-10-24
release: composition@v2.1.4

jeffjakub:

PR #203, opened 2022-10-25
Install Volta w/ `cargo install` to overcome lack of Linux ARM binaries

abernix:
This aims to resolve a blocker encountered by @jeffjakub when trying to land https://github.com/apollographql/federation-rs/pull/202, which appears to be a new transient failure in `volta` and is preventing us from releasing Federation 2.1.4 in both `harmonizer` and `router-bridge`.  The error encountered there is:

```
Releases for non x64 architectures are not currently supported
```

Which is produced by the install script [here](https://github.com/volta-cli/volta/blob/main/dev/unix/volta-install.sh#L129-L136).

Since the Linux ARM binaries are not published and the install script seems to struggle with the request to install from source we need to just build in our more traditional ways using `cargo install`.

We may want to consider pinning the `volta` version if this continues to be problematic, but this should unblock us in the near term. (I did not bother bisecting if there was a previous version on which this worked.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -263,21 +263,25 @@ commands:
                   curl https://get.volta.sh | bash -s -- --skip-setup
                   echo 'export VOLTA_HOME=$HOME/.volta' >> $BASH_ENV
                   echo 'export PATH=$VOLTA_HOME/bin:$PATH' >> $BASH_ENV
-      
+
       - when:
           condition:
             equal: [ *arm_ubuntu_executor, << parameters.platform >> ]
           steps:
             - run:
                 name: Install volta
+                # The install script seems to fail with a message saying:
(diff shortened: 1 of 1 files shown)


garypen (commented):
I note that: https://github.com/volta-cli/volta/releases/download/v1.1.0/volta-1.1.0-macos-aarch64.tar.gz definitely exists, so I'm surprised by your comment about non-x86 architectures.

I just tried following along with the old install and it worked fine on my M1.

What am I missing about this change?

abernix:
I'm talking about linux ARM not macOS ARM.

abernix:
Also worth noting that the passing of the CI tests on this PR are proof that the work-around is working correctly.

garypen (approved):

jeffjakub (commented):
LGTM

PR #212, opened 2022-12-05
Upgrade harmonizer to use typescript

michael-watson:
This PR upgrades the `federation-2/harmonizer` to use typescript instead of javascript.


EverlastingBugstopper (commented):
this is looking good!!! would you mind also updating `federation-1/harmonizer` to do the same thing? the code is structured very similarly, the main difference is that it's pulling from a different upstream library

michael-watson:
Just added federation-1 harmonizer to Typescript, it would be great if you could test it out @EverlastingBugstopper. I got a weird build error when testing, but I don't have a standard config file to test that I know works with Fed 1 (I only have one for Fed 2)

michael-watson:
@EverlastingBugstopper I can't figure out the last test failing. CircleCI is showing an error formatting from code that this PR doesn't touch. I also can't get the same test to fail on my machine.

federation-1/harmonizer/esbuild/bundler.js
```diff
@@ -1,13 +1,33 @@
...
+    build.onResolve({ filter: /^fs$/ }, (args) => ({
+      path: args.path,
+      namespace: "deno-fs",
+    }));
+
+    build.onLoad({ filter: /.*/, namespace: "deno-fs" }, () => ({
+      contents:
+        "module.exports = { existsSync: Deno.ensureFileSync, writeFileSync: Deno.writeTextFile }",
+      loader: "js",
+    }));
```
EverlastingBugstopper, line 14: do we need this for sure? i'm hesitant to add shims for the file system when really the upstream libraries shouldn't depend on these APIs being present. they're just running on inputs/outputs to a function and shouldn't have any side effects like writing to the file system - rover or other library consumers would take care of that bit

federation-2/harmonizer/js-src/composition.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!serviceList || !Array.isArray(serviceList)) {
+    throw new Error("Error in JS-Rust-land: serviceList missing or incorrect.");
+  }
+
+  serviceList.some((service) => {
+    if (
+      typeof service.name !== "string" ||
+      !service.name ||
+      (typeof service.url !== "string" && service.url) ||
+      (typeof service.sdl !== "string" && service.sdl)
```
EverlastingBugstopper, line 16: we probably don't need these runtime checks now that we have typescript involved. what do you think?

federation-2/harmonizer/js-src/test_logger_worker.ts
```diff
@@ -0,0 +1,94 @@
...
+        default:
+          logger.error(`unknown message received: ${JSON.stringify(event)}\n`);
+          break;
+      }
+    } catch (e) {
+      logger.error(`an unknown error occured ${e}\n`);
+    }
+  }
+}
+
```
EverlastingBugstopper, line 93: do we use this file somewhere? can/should it be deleted? i don't see an equivalent in fed-1 code

EverlastingBugstopper (approved):
very close on this one

federation-1/harmonizer/esbuild/bundler.js
michael-watson: You're correct, we don't need this. I pushed a change

PR #221, opened 2023-01-04
Add query planner benchmark + update deno dependency

### LINKED ISSUE #219: Concurrent query planning
We should measure and improve the way our query planner scales.

My theory is that due to the CPU bound nature of planning, the two ways to scale the planner would involve either:
- Spawning several deno runtimes
- Using web workers within a deno runtime

I'm tempted to favor the ~former~ latter approach, for memory footprint reasons. But before we try to improve the performance, we need to profile this.

Steps to take:

- [x] write a benchmark that measures query planning timing and how long planning N queries takes using a single deno runtime
- [x] update deno dependencies to be able to use new features
- [ ] implement and measure the perf gains of webworkers

o0Ignition0o:
Fixes parts 1 and 2 of #219


EverlastingBugstopper (approved):
this is great 👍🏼

PR #232, opened 2023-01-12
create github action for releases

clenfest:
Create two github workflows
    - "Release Components" will create PRs for harmonizer and router-bridge when triggered by the creation of a release in federation
    - "Publish Components" will run the cargo xtask tag command when the release PRs have been merged


clenfest:
Ok, this is starting to look good. Look here for an example of a PR created by this action: https://github.com/apollographql/federation-rs/pull/243

Obviously we'd like to create a pr for the router-bridge as well, but since we need to edit the Cargo.toml file, it's not quite as easy. This action will be triggered inside of the `federation` repo whenever a build is published.

See this comment for how we plan to do it: https://github.com/orgs/community/discussions/26323#discussioncomment-3251454

jeffjakub (approved):

PR #236, opened 2023-01-19
Router-bridge@v2.3.0-beta.3

clenfest:


package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "federation-rs",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
```
abernix, line 6: Is the introduction of this file intentional? And if so, why?

abernix (changes requested):

package-lock.json
clenfest: I think I may have run the npm install from the wrong directory initially. Made this PR very late last night and logged off immediately, thanks for the catch.

abernix (approved):

package-lock.json
abernix: Ah, that checks out.  Thanks!  Approved!

PR #252, opened 2023-02-06
update commit message for automated PRs

clenfest:


abernix:
I've offered this suggestion before but are we sure we don't want to just turn Renovate on for this repository and let it handle the creation of the PRs?  Renovate already does all the dependency management stuff and the PR creation work and there is also off the shelf tooling to do Cargo crate publishing that we could probably use.

Plus, we really should to turn it on anyways for dependency management and when we do it's going to open PRs automatically when new versions of `@apollo/query-planner`, `@apollo/composition`, `@apollo/query-graphs`, `@apollo/federation-internals` are published unless we configure it **not** to.

Plus, it has the added benefit of automatically adding links to the appropriate changelogs, etc., and I think would tell a more complete story?

clenfest:
> I've offered this suggestion before but are we sure we don't want to just turn Renovate on for this repository and let it handle the creation of the PRs? Renovate already does all the dependency management stuff and the PR creation work and there is also off the shelf tooling to do Cargo crate publishing that we could probably use.
> 
> Plus, we really should to turn it on anyways for dependency management and when we do it's going to open PRs automatically when new versions of `@apollo/query-planner`, `@apollo/composition`, `@apollo/query-graphs`, `@apollo/federation-internals` are published unless we configure it **not** to.
> 
> Plus, it has the added benefit of automatically adding links to the appropriate changelogs, etc., and I think would tell a more complete story?

These were very much a mechanism to automate the manual steps that were in our release process in order to save whoever had to do the release time. I'm very open to other approaches if we think it's worth it to make those changes, but this should be a 1 for 1 with what we were doing before. I didn't think about renovate because of the cargo build steps, but I'm happy with that approach if we prefer it.

PR #258, opened 2023-03-01
reuse JsWorker instances

Geal:
deno tends to leak data when creating a runtime, so by reusing a JsWorker instance, we would make sure that it only leaks once instead of on every schema change


Geal:
closing this in favor of #259

PR #266, opened 2023-03-10
Add AbortSignal to global namespace to support lru-cache

clenfest:


federation-2/router-bridge/js-src/types.ts
```diff
@@ -1,3 +1,7 @@
 export type OperationResult =
   | { Ok: any; Err?: undefined }
   | { Ok?: undefined; Err: any };
+
```
trevor-scheer, line 4:
```suggestion
// `lru-cache` (in our dependencies) uses the global `AbortSignal` type
// which isn't readily available in our global types, though it is available
// in deno.
// https://github.com/isaacs/node-lru-cache/pull/247#issuecomment-1204481394
// https://deno.land/api@v1.29.2?s=AbortSignal
```

trevor-scheer (approved):

PR #276, opened 2023-03-17
Make `BuildErrors.build_errors` accessible publicly

xuorig:
~Hey folks, wondering if you would be open to making the `build_errors` field on `BuildErrors` `pub` ? We're using harmonizer for some internal tooling and would like access to individual composition errors.~

Exposing `iter` instead to keep `build_errors` private.

### DIFF
--- a/apollo-federation-types/src/build/error.rs
+++ b/apollo-federation-types/src/build/error.rs
@@ -77,7 +77,7 @@ impl Display for BuildError {
 
 #[derive(Debug, Deserialize, Serialize, Default, Clone, PartialEq, Eq)]
 pub struct BuildErrors {
-    build_errors: Vec<BuildError>,
+    pub build_errors: Vec<BuildError>,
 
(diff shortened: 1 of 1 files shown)


EverlastingBugstopper (approved):

EverlastingBugstopper:
Hey @xuorig - thanks for this contribution, I'd love to get this merged up soon for you so you can get those errors programatically. Would you mind running `cargo fmt --all` on your checkout and pushing those changes up so lints pass? You can then run `cargo xtask lint` to make sure lints will pass by running them on your local machine.

xuorig:
Done @EverlastingBugstopper, unfortunately it looks like CI won't run for my user, but it passes locally for me.

EverlastingBugstopper:
it's unfortunate that CI checks won't run for you, I'm not sure why that is. I checked this out locally and tests and lints pass and this is a very small change. merging this up.

PR #279, opened 2023-03-20
add two tests around defered queries

o0Ignition0o:


federation-2/router-bridge/js-src/plan.ts
```diff
@@ -148,6 +148,7 @@ export class BridgeQueryPlanner {
         },
         errors: [
           {
+            message: e.toString(),
```
abernix, line 151: @o0Ignition0o Is this intentional to leave here?

abernix:
@o0Ignition0o Does this close now that #281 has landed?

o0Ignition0o:
Yes, closing

PR #281, opened 2023-03-21
Fix: router-bridge: expose crypto.getRandomValues() in the deno runtime.

### LINKED ISSUE #280: router-bridge: runtime doesn't have `crypto.getRandomValues`
We have temporarily yanked `0.1.21+v2.4.0` from crates.io as we discovered in a downstream package that it was resulting in `crypto.getRandomValues` errors which are stemming from a `uuid` package's use of that function.  While that function is available in Deno, it isn't available at the known location by default.  Federation 2.4.0 is otherwise seemingly fine, but we'll need to expose that before re-publishing the package.  This came in via the 2.4.0 updates because of a dependency update on `uuid` itself (from 8.x to 9.x, which dropped support for older runtimes and changed the reliance on `crypto`, slightly.)

o0Ignition0o:
Fixes #280 

The npm UUID package v9 relies on [crypto.getRandomValues()](https://github.com/uuidjs/uuid#getrandomvalues-not-supported) which wasn't exposed in the Deno runtime.

This PR exposes getRandomValues via deno_crypto (and its dependencies), and adds a couple of regression tests around defer to make sure we catch this in the future.


federation-2/router-bridge/js-src/plan.ts
```diff
@@ -148,6 +148,7 @@ export class BridgeQueryPlanner {
         },
         errors: [
           {
+            message: e.toString(),
```
o0Ignition0o, line 151: @jesse-apollo WDYT? we can keep this to get more detailed errors in case of a throw that doesn't have a message, but the tradeoff is we might leak implementation detail.

abernix: What sort of implementation detail would be potentially leaked here?

o0Ignition0o: Errors from a call to the query planner `plan()` function (eg: missing crypto.getRandomValues() would make it all the way to the client (in the graphql errors array).

pcmanus (approved):

federation-2/router-bridge/js-src/plan.ts
abernix: Is there a way to get this to the Router log but not to the client?

federation-2/router-bridge/build.rs
```diff
@@ -82,6 +82,13 @@ fn create_snapshot(out_dir: &Path) {
 
     let options = RuntimeOptions {
         will_snapshot: true,
+        extensions: vec![
+            deno_webidl::init(),
+            deno_console::init(),
+            deno_url::init(),
+            deno_web::init::<Permissions>(deno_web::BlobStore::default(), Default::default()),
```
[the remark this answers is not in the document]
o0Ignition0o: Crypto depends on [web](https://github.com/denoland/deno/blob/2fcf1f14cf29bb68995f652f93a4f6e3cb55c8d7/ext/crypto/Cargo.toml#L28) which depends on [bench, console, url, and webidl](https://github.com/denoland/deno/blob/2fcf1f14cf29bb68995f652f93a4f6e3cb55c8d7/ext/web/Cargo.toml#L27)

I wish it wasn't the case...

federation-2/router-bridge/js-src/runtime.js
```diff
@@ -17,6 +17,18 @@ function done(result) {
...
+    Deno.core.ops.op_crypto_get_random_values(arg);
+    return arg;
+  },
+};
+
+// This comes in handy when debugging dependencies in the node modules.
+console = {
+  log: (...args) => print(...args),
+};
+
```
abernix, line 31: Can you speak more to this?  Why not just debugging using `print`?

My appreciation for using `print` — rather than `console.log` — is that some rogue module that starts printing things with `console` will **just fail** rather than _start logging things to the Router's console_

federation-2/router-bridge/src/planner.rs
```diff
@@ -1680,4 +1680,145 @@ feature https://specs.apollo.dev/unsupported-feature/v0.1 is for: SECURITY but i
 
         assert_eq!(expected.to_string(), error_to_display.to_string());
     }
+
+    #[tokio::test]
+    async fn deno_must_support_crypto_get_random_values() {
```
abernix, line 1685: This'll work but it seems like testing it via a bit of indirection.  I wonder if we should just have a small test in `js.rs` that literally just tests the JS runtime and a specific test that literally calls `crypto.getRandomValues`.

At the very least, please put a comment on here that says why we're testing this (e.g., because of a transitive dependency on `crypto.` via `uuid`, via , etc. — otherwise I don't think we'll understand why crypto is even in the picture.

federation-2/router-bridge/js-src/plan.ts
o0Ignition0o: we could log this using the deno logger functions, and redact the messages and only keep the extensions.code, possibly as an error because this specific step (planning) should happen after validation.

federation-2/router-bridge/js-src/runtime.js
o0Ignition0o: That's fair, removing it.

federation-2/router-bridge/build.rs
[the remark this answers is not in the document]
abernix: fine. 😉 

if we regret this, let's replace this polyfill with another.

federation-2/router-bridge/js-src/runtime.js
o0Ignition0o: fixed in 573b40c

federation-2/router-bridge/src/planner.rs
o0Ignition0o: fixed in 4c31b6b

abernix (approved):

federation-2/router-bridge/js-src/plan.ts
o0Ignition0o: Fixed in f4edcf8

o0Ignition0o: Opened https://github.com/apollographql/federation-rs/issues/282

PR #309, opened 2023-04-06
publicize BuildError type and BuildErrorType enum

xuorig:
Sorry to bug you again @EverlastingBugstopper, this is a follow up to: https://github.com/apollographql/federation-rs/pull/276

Can't currently match on the BuildError type.


EverlastingBugstopper:
@xuorig no worries at all! `harmonizer` was thrown together pretty quickly to enable plugins in `rover` specifically so (as you've figured out) not much thought has been given to actually using `harmonizer` as a proper library. thanks for paving this path for yourself (and those who follow 😄 )

EverlastingBugstopper (approved):

PR #312, opened 2023-04-06
Block `centos@8` update

trevor-scheer:
CentOS 8 is EOL, but CentOS 7 is LTS until June 30, 2024. Closes #297 (which is failing CI due to broken support in v8).


trevor-scheer:
We might just prefer to move away from centos altogether if that's an option.
https://github.com/apollographql/federation-rs/pull/302#issuecomment-1499662016

abernix (approved):

PR #318, opened 2023-04-13
feat: Add `introspection_headers` to subgraph config for Rover feature

dbanty:
Supports https://github.com/apollographql/rover/issues/615

BREAKING CHANGE:  A new field was added to SchemaSource::SubgraphIntrospection which is not `#[non_exhaustive]`

Also added some local unit tests for that enum.


EverlastingBugstopper (approved):
this lgtm. tested that this change is backwards compatible w/older versions of rover 👍🏼

PR #321, opened 2023-04-18
Support stdin for supergraph compose

jsegaran:
We don't want to have to create a file every time a build step is run so instead allow using stdin. This still uses the yaml format so probably not what we want longterm for the build steps but was just testing.

### DIFF
--- a/federation-2/supergraph/src/command/compose.rs
+++ b/federation-2/supergraph/src/command/compose.rs
@@ -1,3 +1,5 @@
+use std::io::Read;
+
 use camino::Utf8PathBuf;
 use structopt::StructOpt;
 
@@ -13,7 +15,7 @@ pub struct Compose {
     ///
     /// NOTE: Each subgraph entry MUST contain raw SDL
     /// as the schema source.
-    config_file: Utf8PathBuf,
+    config_file: Option<Utf8PathBuf>,
 }
 
 impl Compose {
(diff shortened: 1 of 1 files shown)


federation-2/supergraph/src/command/compose.rs
```diff
@@ -13,7 +15,7 @@ pub struct Compose {
     ///
     /// NOTE: Each subgraph entry MUST contain raw SDL
     /// as the schema source.
-    config_file: Utf8PathBuf,
+    config_file: Option<Utf8PathBuf>,
```
EverlastingBugstopper, line 18: i would want to design this to require sending `-` explicitly to specify that input is coming from `stdin`. check out `FileDescriptorType` in rover: https://github.com/apollographql/rover/blob/main/src/utils/parsers.rs#L14

The `FromStr` is what lets you specify either `-` or a file: 

```rust
impl FromStr for FileDescriptorType {
    type Err = io::Error;

    fn from_str(input: &str) -> std::result::Result<Self, Self::Err> {
        if input == "-" {
            Ok(FileDescriptorType::Stdin)
        } else if input.is_empty() {
            Err(io::Error::new(
                io::ErrorKind::InvalidInput,
                anyhow!("The file path you specified is an empty string, which is invalid."),
            ))
        } else {
            let path = Utf8PathBuf::from(input);
            Ok(FileDescriptorType::File(path))
        }
    }
}
```

EverlastingBugstopper, line 18: we could add an `input_type: InputType` argument here that is

```rust
enum InputType {
  Yaml
  Json
}
```
and then have methods on it like

```rust
impl InputType {
  fn read_input_from_file(&self, config_input: &FileDescriptorType) -> Result<SupergraphConfig> {
    let contents = config_input.read_file_descriptor()?;
    match &self {
      Self::Yaml => SupergraphConfig::new_from_yaml_file(&contents),
      Self::Json => SupergraphConfig::new_from_json_file(&contents)
    }
  }
}
```

That `new_from_json_file` is new as well, but would look similar to the existing yaml one, it would just call `serde_json::from_str` instead of `serde_yaml::from_str`

make the default be yaml for compatibility w/old versions of rover

```
impl Default for InputType {
  fn default() -> Self {
    Self::Yaml
  }
}
```

federation-2/supergraph/src/command/compose.rs
```diff
@@ -30,10 +32,41 @@ impl Compose {
...
+                        ConfigError::MissingFile {
+                            file_path: config_path.to_string(),
+                            message: e.to_string(),
+                        }
+                    })?;
+                    Ok(contents)
+                } else {
+                    Err(ConfigError::MissingFile {
+                        file_path: config_path.to_string(),
+                        message: "Unable to find supergraph.yaml file".to_owned(),
```
EverlastingBugstopper, line 59: would want to use the real file name (or say it came from stdin) here instead of `supergraph.yaml`

EverlastingBugstopper (commented):
left a few comments - i think allowing json via stdin is a great addition to this interface and i'd love to get it merged (and maybe using it in newer versions of rover)

PR #323, opened 2023-04-18
chore: use large macos executors

### LINKED ISSUE #306: chore: update circleci runners to large m1
this should significantly improve time to release. we already did this for rover, see the [example PR](https://github.com/apollographql/rover/pull/1557).

EverlastingBugstopper:
fixes #306 - use large macos executors


EverlastingBugstopper:
actually this is a horrible idea because v8 bytecode is not going to compile correctly. changing this to be the large x86 executors. this does mean that building for arm is unblocked now.

PR #355, opened 2023-06-13
Add option to disable validation in JS + marker property for validation errors

goto-bus-stop:
Part of https://github.com/apollographql/router/pull/3134

This makes GraphQL validation optional so the router can disable it if it's doing validation in Rust already.

To support running both validations for a while, errors from validation are marked with `.validationError = true`. The router can use that to tell the difference between validation errors and other errors and compare the result with Rust validation. We can remove that later when we are confident in the Rust validation implementation.

When adding the `validationError` property to errors from `api_schema.ts`, I would lose them during serialization between JS and Rust, and I couldn't really figure that out. But the `api_schema` functions are only used during tests in the router as far as I can see, in real operation the schema validation errors also come from the query planner. So it doesn't really matter if we dont have this in the api schema code. 

To do:
- [x] I think it'll be more robust to mark validation errors, instead of marking every other type of error...


federation-2/router-bridge/js-src/api_schema.ts
```diff
@@ -7,12 +7,29 @@ import {
 
 import { buildSupergraphSchema } from "@apollo/federation-internals";
 
-export function apiSchema(sdl: string): ExecutionResult<String> {
+export interface ApiSchemaOptions {
+  graphqlValidation?: boolean;
```
bnjjj, line 11: Should we call it `enabledGraphqlValidation` ? no strong opinion here, just a thought

goto-bus-stop: I think just using the noun is consistent with other options in the project, eg. `autoFragmentization: true` in `QueryPlanOptions`.

o0Ignition0o (approved):

bnjjj (approved):

PR #363, opened 2023-06-26
fix: Allow proper deserialization of integer federation versions

dbanty:
There is currently an inconsistency in the way parsing numbers for `FederationVersion` works, as seen in https://github.com/apollographql/rover/issues/1647 . For some reason, `DeserializeFromStr` allows integers when passed through `serde_yaml::from_str` but _not_ when the YAML has already been parsed into a `serde_yaml::Value`.

This hand-rolled implementation of Deserialize correctly allows integers 0, 1, and 2 to be parsed no matter the source.

### DIFF
--- a/apollo-federation-types/src/config/version.rs
+++ b/apollo-federation-types/src/config/version.rs
@@ -4,8 +4,8 @@ use std::{
 };
 
 use semver::Version;
-use serde::{Deserialize, Deserializer};
 use serde::de::Error;
+use serde::{Deserialize, Deserializer};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
 
 use crate::config::ConfigError;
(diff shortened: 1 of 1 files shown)


apollo-federation-types/src/config/version.rs
```diff
@@ -183,3 +185,100 @@ impl FromStr for FederationVersion {
...
+            where
+                E: Error,
+            {
+                match num {
+                    0 | 1 => Ok(FederationVersion::LatestFedOne),
+                    2 => Ok(FederationVersion::LatestFedTwo),
+                    _ => Err(Error::custom(format!(
+                        "specified version `{}` is not supported",
+                        num
+                    ))),
```
dbanty, line 213: An alternative approach is to convert int to string and use `from_str`, which is more _consistent_, but resource inefficient. Either way is fine with me

EverlastingBugstopper (approved):

PR #364, opened 2023-06-26
ci: Install volta from lock file

dbanty:
[Volta install fails for arm Ubuntu](https://app.circleci.com/pipelines/github/apollographql/federation-rs/1859/workflows/c28ac5bf-bbbb-4186-86d4-9862b75e64ed/jobs/8236) because:

1. Their rust-toolchain sets Rust to 1.63
2. Hashbrown 0.14.0 is allowed by their Cargo.toml
3. Hashbrown 0.14.0 MSRV is 1.64

I think just using the lock file should fix this

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -281,7 +281,7 @@ commands:
                   VOLTA_VERSION=$(curl --silent "https://volta.sh/latest-version")
                   git clone https://github.com/volta-cli/volta --branch "v${VOLTA_VERSION}"
                   cd volta
-                  cargo install --path .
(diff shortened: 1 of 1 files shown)


dbanty:
Also though, maybe the prebuilt binaries work again and we should try that?

EverlastingBugstopper:
they dont have prebuilt binaries for arm which is why we build from source for that target and no others

dbanty:
No prebuilt ARM just yet: https://github.com/volta-cli/volta/issues/1049

PR #369, opened 2023-06-27
Add build plugin types to federation-rs

jsegaran:
This PR is easiest to go commit by commit. It adds json parsing for `SupergraphConfig` since we use that in the plugins. It then adds the json types we use as return values for the plugins.

### DIFF
--- a/apollo-federation-types/src/build_plugin/build_message.rs
+++ b/apollo-federation-types/src/build_plugin/build_message.rs
@@ -1,0 +1,137 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+#[non_exhaustive]
+pub enum BuildMessageLevel {
+    Debug,
+    Info,
+    Warn,
+    Error,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct BuildMessageLocation {
+    line: u32,
+    column: u32,
+
+    #[serde(flatten)]
+    other: crate::UncaughtJson,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct BuildMessage {
+    pub level: BuildMessageLevel,
+    pub message: String,
+    pub step: Option<String>,
+    pub code: Option<String>,
+    pub locations: Vec<BuildMessageLocation>,
+    pub schema_coordinate: Option<String>,
+
+    #[serde(flatten)]
+    other: crate::UncaughtJson,
+}
+
+impl BuildMessage {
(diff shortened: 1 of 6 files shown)


apollo-federation-types/src/build_plugin/plugin_result.rs
```diff
@@ -0,0 +1,157 @@
...
+            Err(json_error) => Self {
+                is_success: false,
+                schema: None,
+                build_messages: BuildMessage::to_build_errors(
+                    vec![format!(
+                        "Could not parse JSON from Rust. Received error {}",
+                        json_error
+                    )],
+                    Some("PLUGIN_EXECUTION".to_string()),
+                    Some("PLUGIN_EXECUTION".to_string()),
```
EverlastingBugstopper, line 37: what's going on here?

apollo-federation-types/src/build_plugin/build_message.rs
```diff
@@ -0,0 +1,137 @@
...
+#[non_exhaustive]
+pub enum BuildMessageLevel {
+    Debug,
+    Info,
+    Warn,
+    Error,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct BuildMessageLocation {
```
EverlastingBugstopper, line 14: should add a comment to all of these structs that any new fields must be wrapped in an `Option` type in order to maintain backwards compatibility. this is to allow new versions of rover to update this struct and still be able to read output from plugins that don't include the optional field. if these structs add a new required field, old plugins won't be able to run anymore because their output won't include that new required field.

EverlastingBugstopper (approved):

apollo-federation-types/src/build_plugin/plugin_result.rs
jsegaran: We use `code` and `step` to highlight display information in studio. I'm trying to indicate that this is a plugin execution error as opposed to a error from the plugin itself.

I was thinking in studio we can build alerting on these kinds of errors as well as give better tool tip information if these appear. 

Ideally this could be an enum, but I don't think we have an exhaustive list of codes or steps anywhere, and we haven't really discussed how we want people to use them programatically.

EverlastingBugstopper: Makes sense - kinda confusing to see both the code and the step be the same exact string.

I think your instinct to make an enum is OK, and honestly I think it's fine that we don't know 100% what it will always be. You can tag enums with `#[non_exhaustive]` in Rust and then all consuming code has to handle the `_` case (something was passed that is not represented by a variant in this enum). I think it would make sense to have two separate enums `Code` and `Step` so you can do `Code::PluginExecution.to_string()` and `Step::PluginExecution.to_string()` and get some sweet readability improvements. You can get the `.to_string()` to work with an `impl Display for Code` and `impl Display for `Step` blocks, VS code should auto-complete the function block for you, but here's an example: https://doc.rust-lang.org/rust-by-example/hello/print/print_display.html

PR #376, opened 2023-07-06
Modify plugin result data types

jsegaran:
### Add non build failure boolean to plugin result
We want to be able to easily differentiate build failures that are related to bugs in calling a plugin from those that are related to non-composing schemas.

We could theoretically use code for this, but we don't have standardization on codes yet. I think adding a new boolean instead will be the best for now, and we can revisit using codes on build messages to indicate this in the future instead.

### Update Message location to match current composition and rover
The original format was based on what the plugins currently output, however we want to support the same set of location based functionality as in rover and composition. This PR updates the locations to match the current rover/composition format, and a subsequent PR will update the plugins to output this format as well.

### Change `BuildMessage::to_build_errors` to a more traditional style constructor in `BuildMessage::new_error`
We never ended up passing a vec of messages instead, we always passed a single one which made this function a bit tedious. The new format is a bit nicer to create adhoc errors with.

### DIFF
--- a/apollo-federation-types/src/build_plugin/plugin_result.rs
+++ b/apollo-federation-types/src/build_plugin/plugin_result.rs
@@ -1,19 +1,27 @@
 use super::BuildMessage;
 use serde::{Deserialize, Serialize};
 
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+#[non_exhaustive]
+// This represents the reason for a build failrue
+pub enum PluginFailureReason {
+    /// If the plugin failed because user inputs can't be built
+    Build,
+    /// If the configuration sent to the plugin is invalid
+    Config,
+    /// If the plugin failed for some internal reason
+    InternalFailure,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 /// PluginResult represents the output of a plugin execution
 /// New fields added to this struct must be optional in order to maintain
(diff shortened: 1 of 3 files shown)


apollo-federation-types/src/build_plugin/plugin_result.rs
```diff
@@ -11,12 +11,42 @@ pub struct PluginResult {
...
     #[serde(flatten)]
     other: crate::UncaughtJson,
 }
 
 impl PluginResult {
+    /// Returns true if the failure is due to something other than the execution of the underlying javascript code
+    /// e.g. the supergraph yaml is misconfigured or the json recieved is malformed
+    pub fn is_non_build_failure(&self) -> bool {
+        self.is_non_build_failure && self.schema.is_none() && !self.is_success
+    }
```
EverlastingBugstopper, line 28: we kinda have this concept in this repo w/the `BuildErrorType` enum:

```
#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
#[serde(rename_all = "lowercase")]
#[non_exhaustive]
pub enum BuildErrorType {
    Composition,
    Config,
}
```

- i like modeling things with enums specifically like this just to make things a bit easier to read. if you stored this type on the struct, this function becomes

```suggestion
    pub fn is_non_build_failure(&self) -> bool {
        !matches!(self, BuildErrorType::Composition)
    }
```

jsegaran: Went with an enum, also using a result to make it more clear we will have a success or error reason. I think it makes the data object a bit easier to understand.

PR #381, opened 2023-07-10
chore: xtask tag now also creates an empty commit

EverlastingBugstopper:
in order to prevent errors where releases fail due to `apollo-federation-types` being published and the tag coexisting at `HEAD` along with other packages, each `cargo xtask tag` invocation will also push an empty commit to `main`.

### DIFF
--- a/xtask/src/tools/git.rs
+++ b/xtask/src/tools/git.rs
@@ -132,6 +132,22 @@ impl GitRunner {
         Ok(())
     }
 
+    // takes a PackageTag and creates an empty commit after publish
+    pub(crate) fn empty_commit(&self, package_tag: &PackageTag, dry_run: bool) -> Result<()> {
+        if !dry_run {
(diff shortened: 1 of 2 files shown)


xtask/src/tools/git.rs
```diff
@@ -132,6 +132,22 @@ impl GitRunner {
         Ok(())
     }
 
+    // takes a PackageTag and creates an empty commit after publish
+    pub(crate) fn empty_commit(&self, package_tag: &PackageTag, dry_run: bool) -> Result<()> {
```
jsegaran, line 136: Does this push to main? If so do we have permissions to push directly to main?

EverlastingBugstopper: ah that's a good point. i looked in the settings and I think it's just me and @abernix that can. perhaps a different strategy is better here.

PR #408, opened 2023-09-13
feat(router-bridge): add validation function to export

lrlna:
This branch is to simply export validation functionality in order for us to do differential testing between js & rust validations. Please do not merge.


federation-2/router-bridge/src/validate.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+pub struct ValidationResponse {
+    /// The introspection response if batch_introspect succeeded
+    #[serde(default)]
+    #[serde(rename = "Ok")]
+    data: Option<serde_json::Value>,
+    /// The errors raised on this specific query if any
+    #[serde(default)]
+    #[serde(rename = "Err")]
+    errors: Option<Vec<ValidationError>>,
```
goto-bus-stop, line 39:
```suggestion
    pub data: Option<serde_json::Value>,
    /// The errors raised on this specific query if any
    #[serde(default)]
    #[serde(rename = "Err")]
    pub errors: Option<Vec<ValidationError>>,
```

I think these need to be pub so apollo-rs fuzzing can read them

PR #419, opened 2023-10-16
Allow harmonize to return a limited number of nodes

tinnou:
## Description

For our largest graph, running composition yields thousands of build hints and each build hint can contain hundreds of attached nodes (nodes are snippets of subgraph SDL that help with locating issues), leading to very high memory consumption (>10sGB) and can use all of the host physical memory.

## Investigation

Running composition using the `harmonize` function with dhat enabled; we can see that most allocations are done when deserializing json from v8 to rust ` serde_v8::de::to_utf8_fast`. On my M1 Mac this was taken right before the process was killed by the OS and was consuming 45GB of memory.

Screenshot 2023-10-16 at 2 17 03 PM

[dhat-heap-main.json](https://github.com/apollographql/federation-rs/files/12922166/dhat-heap-main.json)

## Fix/Proposal

In the spirit of keeping the library backward compatible, I added a `harmonize_limit()` function that takes an extra parameter that will limit the number of nodes returned in each build hint and/or build error. All nodes above that limit are omitted from the result and a flag `omitted_nodes_count` is set so the caller is aware some nodes have been omitted. 

If no `nodes_limit` value is passed it defaults to unlimited. The `harmonize` function defaults to unlimited to preserve backward compatibility.

Passing `nodes_limit=20` works for us for now, but one could argue there is limited additional insight past 20 nodes anyway and we could make it the default instead.

### DIFF
--- a/apollo-federation-types/src/build/error.rs
+++ b/apollo-federation-types/src/build/error.rs
@@ -273,7 +273,7 @@ mod tests {
     use super::{BuildError, BuildErrors};
 
     use serde_json::{json, Value};
-    use crate::build::{BuildErrorNode, BuildHint};
+    use crate::build::{BuildErrorNode};
 
     #[test]
     fn it_supports_iter() {
--- a/apollo-federation-types/src/build/error.rs
+++ b/apollo-federation-types/src/build/error.rs
@@ -6,6 +6,7 @@ use std::{
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
 pub struct BuildError {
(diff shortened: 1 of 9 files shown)


tinnou:
> 👋 @tinnou can you rebase it on latest and also add corresponding changes to `federation-1/harmonizer`?

Rebased main, fixed build & tests.
I did not implement these changes into federation-1 as build hints aren't really a thing in fed v1 side of the harmonizer.

peakematt:
@tinnou @dariuszkuc I see the problem with the failing semgrep job. I'm planning to push a dev version of the job to this branch to validate that my change does fix the issue, then will update to a release version of the job shortly after.

o0Ignition0o (approved):

PR #437, opened 2024-01-04
Enable progressive override in router-bridge

trevor-scheer:
`buildQueryPlan` now accepts an options object currently consisting of an optional Map of overridden labels to enable progressive override. This set of changes will enable the router to propagate these options through to the query planner.

TODO:
- [x] Release v2.7 fed components and update package file to use npm published versions


xtask/src/utils.rs
```diff
@@ -47,8 +47,7 @@ pub(crate) fn get_workspace_roots() -> Result<Vec<Utf8PathBuf>> {
 }
 
 fn project_root() -> Result<Utf8PathBuf> {
-    let manifest_dir = Utf8PathBuf::try_from(MANIFEST_DIR)
-        .with_context(|| "Could not find the root directory.")?;
+    let manifest_dir = Utf8PathBuf::from(MANIFEST_DIR);
```
trevor-scheer, line 50: Seems questionable, but clippy made me do it

o0Ignition0o (approved):

PR #479, opened 2024-05-13
Consume output of `JsRuntime::execute_script` without using Deno ops like `op_composition_result`

benjamn:
After struggling with a bizarre `op_composition_result` serialization regression (CicleCI-only, macOS-only, and `cargo build --release`-only; see #476), I found myself wondering why we need such a complicated system (JS calling a global `done(<result>)` function which calls the Rust `op_composition_result` callback, which then sends data over an IPC channel, serializing it and deserializing it all over again) just to obtain the output of `JsRuntime::execute_result`. Shouldn't `execute_result` be capable of returning a value of some kind, directly to its caller?

In short, the answer to that question is **yes**, and it means we can completely do away with the `done(<result>)` API, along with the underlying Deno ops and the IPC channel system.

This effectively sidesteps the source of the serialization regression, so we don't have to spend any more time diagnosing what exactly went wrong, though I will note this problem appears to affect all previously published versions we might want to rebuild via CircleCI with `cargo build --release --target aarch64-apple-darwin`, so it's not something that changed in our code, and probably not something we can fix without spending a lot more time investigating what changed in the CircleCI build environment.

It's unfortunate this overcomplicated pattern was copied in several places, but they're all slightly different now, so I don't see a good way to unify them.

### DIFF
--- a/federation-1/harmonizer/js-src/composition.ts
+++ b/federation-1/harmonizer/js-src/composition.ts
@@ -73,10 +73,5 @@ function getPosition(token: Token): Position {
 
 //@ts-ignore
 function parseTypedefs(source: string) {
-  try {
-    return parse(source);
-  } catch (err) {
-    // Return the error in a way that we know how to handle it.
-    done({ Err: [{ message: err.toString() }] });
-  }
+  return parse(source);
 }
--- a/federation-1/harmonizer/js-src/do_compose.ts
+++ b/federation-1/harmonizer/js-src/do_compose.ts
@@ -7,17 +7,14 @@ import type { CompositionResult } from "./types";
  * They'll be stripped in the emitting of this file as JS, of course.
  */
 declare let composition_bridge: { composition: typeof composition };
-
-declare let done: (compositionResult: CompositionResult) => void;
 declare let serviceList: { sdl: string; name: string; url: string }[];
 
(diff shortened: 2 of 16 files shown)


dariuszkuc (approved):

benjamn:
Alas, when I reran the `arm_macos` job with SSH and built with
```sh
cd federation-2
cargo build --release --target aarch64-apple-darwin
```
I got [the same strange `serde_v8` error](https://github.com/apollographql/federation-rs/pull/476#issuecomment-2101220771) we've been seeing when using `op_composition_result`:
```sh
$ target/aarch64-apple-darwin/release/supergraph compose repro_ok.yml 
{"Err":{"build_errors":[{"message":"Unable to deserialize composition result: serde_v8 error: invalid type; expected: object, got: string","code":null,"type":"composition","nodes":null,"omittedNodesCount":null}]}}
```
It's somewhat encouraging that the error is now reported as JSON and the process exits, rather than no output and the process hanging, but this suggests the original problem has something to do with `serde_v8` deserialization, rather than the Deno `#[op]` system (though I am still in favor of the simplification of this PR).

Unfortunately, we'll have to keep debugging this, since publishing these changes in an official release probably would not produce a usable `supergraph` binary.

clenfest (approved):

PR #481, opened 2024-05-14
Update `deno_core` to 0.280.0 (latest) in `federation-2/harmonizer`

benjamn:
This is a big upgrade, and we should probably apply a similar upgrade in `router-bridge` (though it will be trickier there), but the `harmonizer`'s usage of `deno_core` is limited/constrained enough that I think we can be confident this upgrade is safe as long as tests pass.

More immediately, updating `deno_core` will finally fix the `deno_core::serde_v8` deserialization bug I described in https://github.com/apollographql/federation-rs/pull/479#issuecomment-2110342100, which has been mysteriously manifesting only on CircleCI, and only when building with `cargo build --release --target aarch64-apple-darwin`.

I verified this update fixes the problem by SSHing into the CircleCI build machine and building the `supergraph` binary there, and I will re-verify it once the CircleCI checks for this PR complete.


o0Ignition0o (approved):
Awesome! 🚀

benjamn:
Alright, I'm thrilled to report we have solved [the CircleCI build problem](https://github.com/apollographql/federation-rs/pull/476#issuecomment-2101220771) in two independent and robust ways:
* PR #483 pinned the `rustc` version to 1.76.0, instead of the latest `stable` release (1.78.0), which is enough by itself to prevent the `serde_v8` bug in `--release` builds. The sudden `rustc` update appears to have been the original trigger for this bug, since `rustup` just installs the latest `stable` release, whatever it may currently be. That won't happen any more, which is good.
* This PR updates `deno_core` (and by extension `serde_v8`) to the latest version (all the way from 0.200.0 to 0.280.0), which no longer suffers from whatever breaking change shipped in `rustc` 1.78.0, which means we _could_ safely update `rustc` in the future, without reintroducing the bug.

By SSHing into the `arm_macos` CircleCI build machine, I have verified that both `rustc` versions produce a working `supergraph` binary using the changes from this PR (which includes #483).

While simply reverting to `rustc` 1.76.0 would have been the simplest fix for this problem, defense in depth and future-proofing seem worthwhile nevertheless.

dariuszkuc (approved):

PR #482, opened 2024-05-14
Use `deno_core::serde_v8::from_v8` only to obtain `serde_json::Value`

benjamn:
This change tests the hypothesis that `deno_core::serde_v8::from_v8` may not be capable of directly converting to a `Result<BuildOutput, Vec<CompositionError>>`, but instead can give us a generic `serde_json::Value` that `serde_json::from_value` can then convert to the desired `Result` type.


benjamn:
This produced the same so-far-inexplicable `serde_v8 error: invalid type; expected: object, got: string` error, so I guess this hypothesis has been rejected.

PR #488, opened 2024-05-16
Avoid deleting all Git tags before pushing tags

benjamn:
I don't understand why the original author of this code thought it was important to delete all local tags, but that process now takes a considerable amount of time (to delete 359 tags) and is unnecessary. We can just push specific tags instead of using the `git push --tags` command.

The `git pull` command requires a remote tracking branch to be set. Since this is not a guaranteed condition, it's better to skip the unnecessary `git pull` command and assume we're already up-to-date before the `tag_release` function is called.

### DIFF
--- a/xtask/src/tools/git.rs
+++ b/xtask/src/tools/git.rs
@@ -48,19 +48,6 @@ impl GitRunner {
         self.exec(&["fetch", "--tags", "--force"])
     }
 
-    // updates the remote tags we know about,
-    // overwriting any local tags,
-    // and then returns all current local git tags
-    pub(crate) fn get_tags(&self) -> Result<Vec<String>> {
-        self.fetch_remote_tags()?;
-        Ok(self
-            .exec(&["tag"])?
-            .stdout
-            .lines()
-            .map(|s| s.to_string())
-            .collect())
-    }
-
     // gets the current tags that point to HEAD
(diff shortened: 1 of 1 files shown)


xtask/src/tools/git.rs
```diff
@@ -106,20 +93,17 @@ impl GitRunner {
...
-            for local_tag in self.get_tags()? {
-                self.exec(&["tag", "-d", &local_tag])?;
-            }
             // create all the git tags we need from the PackageTag
             for tag in package_tag.all_tags() {
                 self.exec(&["tag", "-a", &tag, "-m", &tag]).context("If you want to re-publish this version, first delete the tag in GitHub at https://github.com/apollographql/federation-rs/current_git_tags")?;
             }
             // push up _only_ the tags that we just created
-            self.exec(&["push", "--tags", "--no-verify"])?;
+            for tag in package_tag.all_tags() {
```
dylan-apollo, line 102: Is there a reason to create them all before pushing any? Or should we merge this loop with the one above it?

dylan-apollo (approved):
Looks good to me, just one question in review. Did you already test this out? If not, I'm happy to kick off another test build from my branch just to verify.

xtask/src/tools/git.rs
benjamn: 99b17f4aca0c35d381d2b3bc9af51928c899cfa1

PR #491, opened 2024-05-21
Remove Fed 1 from main

dylan-apollo:
This simplifies the repo into a single Cargo workspace which removes a bunch of extra infrastructure. The idea is we can have a `v1` or `support/v1` branch where we can still make tweaks to fed 1 (e.g., security fixes) without it being in the way for new fed 2 work.


lennyburdette (approved):

benjamn (approved):
LGTM though I recommend publishing a prerelease to test with Rover (your idea, just agreeing)

dylan-apollo:
Tested by building `2.8.0-connectors.5` with a derivative of this branch (which will become the new connectors branch)

PR #493, opened 2024-05-22
Add amd linux executor

o0Ignition0o:
Add an amd linux executor that matches the apollographql/router's CI environment.

This also reverts #481 until we can find a version of deno_core that works as expected on the amd_linux executor.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -219,6 +219,8 @@ executors:
     docker:
       - image: cimg/base:stable
     resource_class: xlarge
+    environment:
+      XTASK_TARGET: "x86_64-unknown-linux-gnu"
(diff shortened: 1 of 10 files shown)


benjamn (approved):
I'm sorry to see #481 reverted, but we know `deno_core` 0.200.0 worked without issues for a long time, until the accidental `rustc` update from 1.76 to 1.78 introduced the `harmonizer` regression, so I agree this is probably the safest solution for now.

I would be more motivated to make the `deno_core` upgrade stick if we were planning to support the `harmonizer` and `router-bridge` system for a long time, but hopefully their days are numbered, and prioritizing stability is worthwhile until then.

At least we were able to simplify the `harmonizer` system with #479, which we are not reverting at this time.

PR #510, opened 2024-06-19
remove deno_crypto

Geal:
### Motivation

We are responding to https://rustsec.org/advisories/RUSTSEC-2024-0344.  While our usage is not impacted by this (as our "crypto" usage doesn't even come close), but we'd rather not have the dependency in the tree and not updated when there is a clear-enough path around it.

### Practically speaking

Our `deno_crypto` usage was at the root of the imported dependency.  As it is only used for its implementation of `crypto.getRandomValues` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues)), we can easily implement [getting random values](https://docs.rs/deno_crypto/latest/src/deno_crypto/lib.rs.html#140) locally, directly, with the rand crate.


router-bridge/src/worker.rs
```diff
@@ -278,6 +281,26 @@ async fn receive(state: Rc<RefCell<OpState>>) -> Result<JsonPayload, anyhow::Err
...
+    if let Some(seeded_rng) = maybe_seeded_rng {
+        seeded_rng.fill(out);
+    } else {
+        let mut rng = thread_rng();
+        rng.fill(out);
+    }
+
+    Ok(())
+}
+
```
abernix, line 303: Can we leverage any unit tests from the other implementation as well?  I know that "this will be tested somewhere indirectly", but it's not merely intuitive to believe it **is** tested here without the test next to it.

Geal: we have a test in https://github.com/apollographql/federation-rs/blob/main/router-bridge/js-src/test_get_random_values.ts and I checked locally that it worked as expected (function properly exported, etc)

abernix: Ok, just add a comment to the TS plz.

Geal: done

abernix (approved):

PR #511, opened 2024-06-19
fix router-bridge release action

Geal:


abernix (approved):
Well, the current version definitely [isn't](https://github.com/apollographql/federation-rs/actions/runs/9579162830/job/26411018892#step:6:450) working.

PR #512, opened 2024-06-19
allow router-bridge releases independently of federation updates

Geal:
We sometimes need to publish a new router-bridge version without a federation upgrade, so the release action should not fail if we keep the same federation version


abernix (approved):
Agree.   And practically speaking, this value is merely a value that lives inside the `package.json` for a moment while it's built.  I don't believe it could possibly be load-bearing.

PR #516, opened 2024-06-22
Add the capability to increase harmonizer heap size

tinnou:
## Description 
For large to very graphs, it is possible for deno v8 to run out of memory with the default settings. For us, it happens around `1400MB` of heap (I wasn't able to find the default v8 heap size)

Similar to the query planner, this PR vows to allow the harmonizer to grow up to a specified limit using env variables.

The error:
```
    <--- Last few GCs --->
    
    [48487:0x135078000]    51001 ms: Scavenge 1396.0 (1426.0) -> 1394.0 (1426.0) MB, 1.25 / 0.00 ms  (average mu = 0.211, current mu = 0.205) allocation failure; 
    [48487:0x135078000]    51004 ms: Scavenge 1396.0 (1426.0) -> 1394.1 (1426.0) MB, 1.21 / 0.00 ms  (average mu = 0.211, current mu = 0.205) allocation failure; 
    [48487:0x135078000]    51006 ms: Scavenge 1396.0 (1426.0) -> 1394.0 (1430.0) MB, 1.12 / 0.00 ms  (average mu = 0.211, current mu = 0.205) allocation failure; 
    
    
    <--- JS stacktrace --->
    
    
    #
    # Fatal JavaScript out of memory: Ineffective mark-compacts near heap limit
    #
    
```

### DIFF
--- a/harmonizer/Cargo.toml
+++ b/harmonizer/Cargo.toml
@@ -25,6 +25,7 @@ apollo-federation-types = { version = "0.11.0", path = "../apollo-federation-typ
 deno_core = "0.200.0"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+tracing = "0.1.33"
 
 [dev-dependencies]
 insta = "1.34.0"
--- a/harmonizer/src/lib.rs
+++ b/harmonizer/src/lib.rs
@@ -39,6 +39,12 @@ use apollo_federation_types::build::{
     BuildError, BuildErrors, BuildOutput, BuildResult, SubgraphDefinition,
 };
 
+// A reasonable default starting limit for our deno heap.
+const APOLLO_ROUTER_HARMONIZER_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE_DEFAULT: &str = "256";
+// A reasonable default max limit for our deno heap.
+const APOLLO_ROUTER_HARMONIZER_EXPERIMENTAL_V8_MAX_HEAP_SIZE_DEFAULT: &str = "1400";
+
+
 /// The `harmonize` function receives a [`Vec<SubgraphDefinition>`] and invokes JavaScript
 /// composition on it, either returning the successful output, or a list of error messages.
 pub fn harmonize(subgraph_definitions: Vec<SubgraphDefinition>) -> BuildResult {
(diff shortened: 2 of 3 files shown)


harmonizer/src/lib.rs
```diff
@@ -55,13 +61,59 @@ pub fn harmonize_limit(
...
 
+    let initial_heap_size =
+        std::env::var("APOLLO_ROUTER_HARMONIZER_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE").unwrap_or_else(|_e| {
+            APOLLO_ROUTER_HARMONIZER_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE_DEFAULT.to_string()
+        });
+
+    let max_heap_size =
+        std::env::var("APOLLO_ROUTER_HARMONIZER_EXPERIMENTAL_V8_MAX_HEAP_SIZE").unwrap_or_else(|_e| {
+            APOLLO_ROUTER_HARMONIZER_EXPERIMENTAL_V8_MAX_HEAP_SIZE_DEFAULT.to_string()
+        });
```
abernix, line 72: This doesn't run in Router-land, so we should remove `_ROUTER` from these and just go with `APOLLO_HARMONIZER_EXPERIMENTAL_V8_MAX_HEAP_SIZE`

harmonizer/Cargo.toml
```diff
@@ -25,6 +25,7 @@ apollo-federation-types = { version = "0.11.0", path = "../apollo-federation-typ
 deno_core = "0.200.0"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+tracing = "0.1.33"
```
abernix, line 28: I am ... bearish ... on introducing `tracing` to this _just_ for the info logs below if we didn't already have something like that in this package, but i'll see if the team has any other suggestions.

harmonizer/src/lib.rs
tinnou: sounds good, I actually ran some additional tests and it looks like `--max-heap-size` becomes useless as soon as resizing is involved, i.e it's possible to go beyond max-heap-size through resizing. I'm thinking to update the logic so heap resizing only happens if max-heap-size env variable is not provided.

harmonizer/Cargo.toml
tinnou: Ok let me know, happy to remove. Since it was in the bridge I added it but I see as well the argument against.

harmonizer/src/lib.rs
tinnou: Updated the PR to only resize if max heap size has been explicitly set

harmonizer/src/lib.rs
```diff
@@ -55,13 +61,64 @@ pub fn harmonize_limit(
     // The snapshot is created in the build_harmonizer.rs script and included in our binary image
     let buffer = include_bytes!(concat!(env!("OUT_DIR"), "/composition.snap"));
 
+    let initial_heap_size =
+        std::env::var("APOLLO_HARMONIZER_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE").unwrap_or_else(|_e| {
```
abernix, line 65: Drive-by, I _think_ that this unused argument made the compiler upset and thus the failing CI tests. 😄  

```suggestion
        std::env::var("APOLLO_HARMONIZER_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE").unwrap_or_else(|| {
```

tinnou: ah yes thank you, fixed it in a previous commit.

abernix (approved):
LGTM. Also tested locally that it works.

PR #521, opened 2024-06-27
Default `CompositionHint.nodes` to an empty `Vec` when `undefined`

trevor-scheer:
Follow-up to #519


dylan-apollo (approved):
TY! Should this be updated in the TypeScript code to be more clear? And this doesn't return `null`, right, just `undefined`?

PR #532, opened 2024-07-09
temporarily disable intel macos architecture

clenfest:
This PR should be reverted once beta is released


clenfest:
I don't think there's anything I need to do in the code itself to make sure `--real-publish` will work? Can someone confirm?

dariuszkuc (approved):
Approved - but instead of deleting should we just comment those out?

clenfest:
> Approved - but instead of deleting should we just comment those out?

Started out doing that, but I don't think there's a way to comment out part of a line in YAML. I'll revert it later today after I do the release.

PR #533, opened 2024-07-10
Add function to check for macOS support in Federation Versions

jonathanrainer:
At present we have a check for ARM Linux support but no corresponding check for ARM on macOS. This adds that check and a test to ensure we retain the supported versions we want in the future.


dylan-apollo (approved):
LGTM, thanks!

PR #538, opened 2024-07-12
Create github action and include in Circle build pipeline

clenfest:
Build and test x86 mac architecture through a github action

### DIFF
--- a/.github/workflows/tests-mac-x86.yml
+++ b/.github/workflows/tests-mac-x86.yml
@@ -1,0 +1,69 @@
+on:
+  workflow_dispatch:
+    inputs:
+      composition-versions:
+        description: 'JSON list of supergraph versions'
+        required: true
+        type: string
+
+jobs:
+  build-rover:
+    name: Build Rover for macOS x86-64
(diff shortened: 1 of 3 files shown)


.github/workflows/tests-mac-x86.yml
```diff
@@ -0,0 +1,54 @@
...
+  comopse-supergraph:
+    needs: build-supergraph
+    name: Run supergraph compose on macOS x86-64
+    # strategy:
+    #   matrix:
+    #     composition-version: ${{ fromJSON(inputs.composition-versions) }}
+    # x86-64 runner
+    runs-on: macos-13
+    env:
+      TEST_COMPOSITION_VERSION: "=${{ matrix.composition-version }}"
```
duckki, line 42: These env. vars. are not set after all. And, it works fine without them.
Do we still need these lines (and the commented-out `strategy` section up there)?

duckki (approved):
I confirmed the x86 builds ran on this PR branch and I could download the artifact and run it locally. I just have a minor comment on the unused environment variables. Otherwise, it looks good to me!

PR #543, opened 2024-07-16
Move x86 Linux Compilation to `manylinux` container for continued `glibc` 2.17 support

jonathanrainer:
As per title, copying an approach we used for Rover after the CentOS EOL.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ workflows:
           name: Lint
           matrix:
             parameters:
-              platform: [amd_centos]
+              platform: [amd_manylinux]
               rust_channel: [stable]
               command: [lint]
   test:
@@ -53,7 +53,7 @@ workflows:
           name: Run cargo tests (<< matrix.rust_channel >> rust on << matrix.platform >>)
           matrix:
             parameters:
-              platform: [amd_centos, arm_ubuntu, amd_linux, amd_macos, arm_macos, amd_windows]
+              platform: [amd_manylinux, arm_ubuntu, amd_linux, amd_macos, arm_macos, amd_windows]
               rust_channel: [stable]
               command: [test]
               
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -235,6 +235,14 @@ executors:
     environment:
       XTASK_TARGET: "x86_64-pc-windows-msvc"
 
+  amd_manylinux: &amd_manylinux_executor
+    docker:
+      # This image is used for building Python Wheels but it has what we need for Rust compilation and crucially
+      # glibc 2.17.
+      - image: quay.io/pypa/manylinux2014_x86_64:2024.07.02-0
```
tninesling, line 242: Would you be able to share some context on how this image was selected? My initial reaction is that it obfuscates the platform we're testing against (maybe there's an older version of the ubuntu image we can use to get the same effect?).

Second, is `glibc 2.17` significant in some way? I don't think Atlas has been tracking specific dependencies like that, so it'd be helpful to get some context on that and if there are potentially other toolchains we need to target.

jonathanrainer: Hi @tninesling, so I'll attempt to give a potted history where I can:

- For a long time `rover` has created binaries that are compiled with `glibc` 2.17 mostly because we have customers who are running `rover` in very old Enterprise systems that do not want to upgrade. As far as we know this is still a requirement and so we produce and crucially test `rover` binaries in an environment that mirrors this
- Furthermore, because of the use of Deno inside the `supergraph` binary there is some requirement for glibc 2.17 though I'll admit to not quite understanding exactly what that is.
   - If it were possible to start support `musl` builds of `supergraph` we could start to move towards a world where this is irrelevant but that might be challenging today.
- Since `supergraph` v2.8.3 our tests have been failing when we run them in this environment, and the only significant change to the repo has been the change to compiling on a system with newer versions of `glibc`. See https://github.com/apollographql/rover/pull/1983 as an example.
- As such restoring the position where we compile for `glibc` 2.17 should put us back in a good position and allow us to support newer `supergraph` versions as we progress.

I also think there's a wider discussion here about what binaries we support and how we build them, the `router` team have a similar issue and it all comes together in `rover` because it utilises both binaries to make that all work.

Does all that make sense? Happy to have a F2F meeting if it needs more explanation! :)

tninesling: Thanks for the context! My only concern is that this image obfuscates the actual distro that's being used, but if it works for you and covers the `glibc` dependency, I think we can move forward with this.

tninesling (approved):

PR #545, opened 2024-07-19
Send payload id if applicable when an exception is caught

o0Ignition0o:
If an exception is caught during planning, we previously didn't send the messageId.

The response could thus not be properly routed back to the proper sender, and would lead to such errors as:

```
value: TypeError: Error parsing args at position 1: missing field `id`
```

This changeset adds the messageId if possible.

### DIFF
--- a/router-bridge/js-src/plan_worker.ts
+++ b/router-bridge/js-src/plan_worker.ts
@@ -245,8 +245,14 @@ const updateQueryPlanner = (
 
 async function run() {
   while (true) {
+    // If an Exception is raised,
+    // there won't be any messageId to send a payload back to the router,
+    // unless we keep it in this scope.
+    let messageId;
     try {
+      messageId = "";
       const { id, payload: event } = await receive();
+      messageId = id;
       try {
         switch (event?.kind) {
           case PlannerEventKind.UpdateSchema:
@@ -336,6 +342,7 @@ async function run() {
       };
 
       await send({
+        id: messageId,
         payload: {
           errors: [unexpectedError],
           usageReporting: {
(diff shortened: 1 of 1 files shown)


router-bridge/js-src/plan_worker.ts
```diff
@@ -245,8 +245,14 @@ const updateQueryPlanner = (
 
 async function run() {
   while (true) {
+    // If an Exception is raised,
+    // there won't be any messageId to send a payload back to the router,
+    // unless we keep it in this scope.
+    let messageId;
     try {
+      messageId = "";
```
goto-bus-stop, line 253:
```suggestion
    let messageId = "";
    try {
```

goto-bus-stop: same thing but more compact

goto-bus-stop (commented):
Hmm. <strike>It looks like this could only occur if `receive()` errored? I'm not clear on when that would happen (though this patch would surely help to figure that out 😄 )</strike>

I wonder if we should accept `Option<String>` on the rust side rather than using a default empty value? Or is that super invasive?

goto-bus-stop (approved):
Ohh wait. I reread it after. My comment about `receive()` is incorrect.

This patch looks good, we would only get the default empty value if `receive()` fails, which should not happen ( 😇 ), and we now get the correct value if QP fails, which we previously didn't.

Adjusting my question: is getting an empty value going to be a problem in the router? It shouldn't happen, but it might ;p Either way this is better than before.

duckki:
This seems working. I don't see this error with this PR.

PR #557, opened 2024-07-30
impl `From<BuildError>` and `From<BuildHint>` for `Issue`

trevor-scheer:
LSP in Rover needs composition issues in the form of `Issue`s. These implementations allow for easy translation from `supergraph` output into what the language server expects.

### DIFF
--- a/apollo-composition/src/lib.rs
+++ b/apollo-composition/src/lib.rs
@@ -10,6 +10,7 @@ use apollo_federation_types::build_plugin::{
 use apollo_federation_types::javascript::{
     CompositionHint, GraphQLError, SatisfiabilityResult, SubgraphASTNode, SubgraphDefinition,
 };
+use apollo_federation_types::rover::{BuildError, BuildHint};
 use either::Either;
 use std::iter::once;
 use std::ops::Range;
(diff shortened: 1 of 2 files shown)


apollo-composition/src/lib.rs
```diff
@@ -355,3 +356,58 @@ fn satisfiability_result_into_issues(
         Err(issue) => Either::Right(once(issue)),
     }
 }
+
+impl From<BuildError> for Issue {
+    fn from(error: BuildError) -> Issue {
+        Issue {
+            code: error.code.expect("BuildError.code unexpectedly missing"),
```
goto-bus-stop, line 363: `.expect` is for things that never happen. Does this never happen? If it doesn't, should BuildError be changed to not use `Option`s?

apollo-composition/src/lib.rs
```diff
@@ -355,3 +356,58 @@ fn satisfiability_result_into_issues(
...
+                start: LineColumn {
+                    line: start.line.unwrap_or(0),
+                    column: start.column.unwrap_or(0),
+                },
+                end: LineColumn {
+                    line: end.line.unwrap_or(0),
+                    column: end.column.unwrap_or(0),
+                },
+            },
+        }
```
dylan-apollo, line 411: This should return `Option<SubgraphLocation>` and use `?` a bunch instead of defaults. See the way `from_ast` works.

apollo-composition/src/lib.rs
dylan-apollo: Since all these things come from TypeScript, and we have to support arbitrarily old versions, we can't be sure they'll never be null. `unwrap_or_default()` here would be better here. Message should probably have a default string, like "Unknown error" or something. Locations should be able to `into_iter()` on the option to end up with an empty vec if it was null.

dylan-apollo (changes requested):

apollo-composition/src/lib.rs
trevor-scheer: I went ahead and made this change. Is it weird to filter out locations where the linecol info is missing (as opposed to defaulting to 0,0)?

apollo-composition/src/lib.rs
```diff
@@ -355,3 +374,35 @@ fn satisfiability_result_into_issues(
         Err(issue) => Either::Right(once(issue)),
     }
 }
+
+impl From<BuildError> for Issue {
+    fn from(error: BuildError) -> Issue {
+        Issue {
+            code: error.code.unwrap_or("UNKNOWN_ERROR_CODE".to_string()),
+            message: error.message.unwrap_or("Unknown error".to_string()),
```
dylan-apollo, line 382:
```suggestion
            code: error.code.unwrap_or_else(|| "UNKNOWN_ERROR_CODE".to_string()),
            message: error.message.unwrap_or_else(|| "Unknown error".to_string()),
```

To avoid allocating those strings unless the error was actually missing details

apollo-composition/src/lib.rs
```diff
@@ -355,3 +374,35 @@ fn satisfiability_result_into_issues(
...
+                .collect(),
+            severity: Severity::Error,
+        }
+    }
+}
+
+impl From<BuildHint> for Issue {
+    fn from(hint: BuildHint) -> Issue {
+        Issue {
+            code: hint.code.unwrap_or("UNKNOWN_HINT_CODE".to_string()),
```
dylan-apollo, line 397:
```suggestion
            code: hint.code.unwrap_or_else(|| "UNKNOWN_HINT_CODE".to_string()),
```

dylan-apollo (changes requested):

dylan-apollo (approved):
LGTM, thanks!

apollo-composition/src/lib.rs
dylan-apollo: Well setting the location to an arbitrary number doesn't feel right (it'd be 1, 1, not 0, 0)—we should semantically encode that there is no location with `None`. I _could_ see making the entire `range` optional, though, so we can keep subgraph name if it's set.

Actually, I think it doesn't make sense to produce locations _without_ subgraph names, so we should probably filter out any that are missing it.

apollo-composition/src/lib.rs
```diff
@@ -355,3 +356,57 @@ fn satisfiability_result_into_issues(
...
+                    start: LineColumn {
+                        line: start.line.unwrap_or_default(),
+                        column: start.column.unwrap_or_default(),
+                    },
+                    end: LineColumn {
+                        line: end.line.unwrap_or_default(),
+                        column: end.column.unwrap_or_default(),
+                    },
+                })
+            }),
```
dylan-apollo, line 409:
```suggestion
            range: location.start.and_then(|start| {
                let end = location.end?;
                Some(Range {
                    start: LineColumn {
                        line: start.line?,
                        column: start.column?,
                    },
                    end: LineColumn {
                        line: end.line?,
                        column: end.column?,
                    },
                })
            }),
```

dylan-apollo (approved):

PR #558, opened 2024-08-06
get heap statistics from deno

Geal:
This adds a planner event to request V8 heap statistics.

This data is gathered from the JS side because the related methods from the rust side require a `&mut` access to the runtime, which is impossible since it already runs in an infinite loop.
Since we elected to not run the prelude parts of deno which register functions like `Deno.memory_usage`, we intern the [op_runtime_memory_usage function](https://github.com/denoland/deno/blob/897159dc6e1b2319cf2f5f09d8d6cecc0d3175fa/runtime/ops/os/mod.rs#L329) and import it directly as a host function, as we've done already with  `op_crypto_get_random_values`. To avoid hosting too much code, the `rss` field is not filled (it needs platform specific functions), since that info can generally be obtained by other means.

### DIFF
--- a/router-bridge/src/planner.rs
+++ b/router-bridge/src/planner.rs
@@ -400,10 +400,13 @@ where
 
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
+/// deno's heap statistics
 pub struct HeapStatistics {
-    //pub rss: u64,
+    /// total size of the heap for V8, in bytes
     pub heap_total: u64,
+    /// amount of the heap used for V8, in bytes
     pub heap_used: u64,
+    /// emory, in bytes, associated with JavaScript objects outside of the JavaScript isolate
     pub external: u64,
 }
 
--- a/router-bridge/src/planner.rs
+++ b/router-bridge/src/planner.rs
@@ -787,10 +787,9 @@ pub struct QueryPlannerDebugConfig {
 }
 #[cfg(test)]
 mod tests {
+    use futures::stream;
     use futures::stream::StreamExt;
-    use futures::stream::{self};
 
-    use core::panic;
     use std::collections::BTreeMap;
 
     use super::*;
@@ -1904,7 +1903,6 @@ ofType {
         let statistics = planner.get_heap_statistics().await.unwrap();
 
         println!("statistics: {statistics:?}");
-        panic!()
     }
 }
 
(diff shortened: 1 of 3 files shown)


router-bridge/src/planner.rs
```diff
@@ -1871,6 +1891,19 @@ ofType {
...
+    async fn heap_statistics() {
+        let planner =
+            Planner::<serde_json::Value>::new(SCHEMA.to_string(), QueryPlannerConfig::default())
+                .await
+                .unwrap();
+
+        let _subgraphs = planner.subgraphs().await.unwrap();
+        let statistics = planner.get_heap_statistics().await.unwrap();
+
+        println!("statistics: {statistics:?}");
```
tninesling, line 1905: Does this need an assertion, or is the point just that the call above shouldn't panic on `.unwrap()`?

Geal: there's no guarantee of deterministic heap usage, I can't assert on the returned value, so the only thing we really need to know is if the function returned a value

tninesling (approved):

o0Ignition0o (approved):

router-bridge/src/worker.rs
```diff
@@ -302,6 +303,30 @@ fn op_crypto_get_random_values(state: &mut OpState, out: &mut [u8]) -> Result<()
     Ok(())
 }
 
+// HeapStats stores values from a isolate.get_heap_statistics() call
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct MemoryUsage {
+    //rss: usize,
```
abernix, line 310: keep or go? (i commented on the release PR just now but it's the same question)

router-bridge/src/worker.rs
```diff
@@ -302,6 +303,30 @@ fn op_crypto_get_random_values(state: &mut OpState, out: &mut [u8]) -> Result<()
...
+}
+
+// from https://github.com/denoland/deno/blob/897159dc6e1b2319cf2f5f09d8d6cecc0d3175fa/runtime/ops/os/mod.rs#L329
+// tested in planner.rs
+#[op(v8)]
+fn op_runtime_memory_usage(scope: &mut v8::HandleScope<'_>) -> MemoryUsage {
+    let mut s = v8::HeapStatistics::default();
+    scope.get_heap_statistics(&mut s);
+    MemoryUsage {
+        //rss: rss(),
```
abernix, line 323: same.

PR #562, opened 2024-08-12
Build MUSL binaries as well as `glibc` ones

jonathanrainer:
Seems as though we can build `musl` binaries if we want too, though it requires compiling V8 at source, this PR is an experiment to see if that's possible.

In terms of testing I've done the following:
1. Spun up a `cimg/base:stable` Docker image, installed the right tools and built the binary using the same xtask command as is used in CI - ✅ 
2. Installed `rover` in the same Docker image, moved the built `supergraph` binary into the `.rover/bin` directory and run `supergraph compose` against the `retail-supergraph` example - ✅ 
3. Re-done 1. without the `--debug` flag to ensure that we can get a release binary ✅ 

I'm happy to do anymore testing we think is a good idea, because supporting a new binary is a big deal so I want to make sure we're ready if we decide to go ahead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -267,7 +267,7 @@ executors:
     environment:
       XTASK_TARGET: "x86_64-unknown-linux-musl"
       V8_FROM_SOURCE: "1"
-      GN_ARGS: "is_clang=false custom_toolchain=\"x86_64-linux-musl-gcc\""
+      CLANG_BASE_PATH: "/usr/bin"
 
 
 # reusable command snippets can be referred to in any `steps` object
@@ -326,7 +326,7 @@ commands:
           steps:
             - run:
                 name: Install Alpine required Dependencies
-                command: apk add curl bash nodejs npm cmake g++ make python3 gn git gcc glib-dev icu-dev libstdc++ linux-headers ninja tar xz
+                command: apk add curl bash nodejs npm cmake g++ make python3 gn git gcc glib-dev icu-dev libstdc++ linux-headers ninja tar xz musl-dev
 
       - install_rust_toolchain:
           rust_channel: << parameters.rust_channel >>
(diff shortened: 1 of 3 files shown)


lrlna:
Oh man I remember us having conversation about how we could no longer support musl ([more specifically Alpine](https://github.com/apollographql/rover/issues/546)) after introducing JS (and v8) to `rover` from ...maybe 3 years ago. Is the idea to get this musl binary built in `federation-rs` directly so we don't have to build v8 from source higher upstream in `rover` and `router`?

jonathanrainer:
> Oh man I remember us having conversation about how we could no longer support musl ([more specifically Alpine](https://github.com/apollographql/rover/issues/546)) after introducing JS (and v8) to `rover` from ...maybe 3 years ago. Is the idea to get this musl binary built in `federation-rs` directly so we don't have to build v8 from source higher upstream in `rover` and `router`?

Pretty much yes, this PR was more about me testing assumptions that we couldn't build `musl` binaries because of v8 (which you can you just have to compile it yourself). If we do this then we remove one of the bricks in the wall to letting `rover` support composition under `musl` binaries.

lrlna:
> If we do this then we remove one of the bricks in the wall to letting rover support composition under musl binaries.

what are the implications to upstream dependencies like rover and router with a change like this?

jonathanrainer:
> > If we do this then we remove one of the bricks in the wall to letting rover support composition under musl binaries.
> 
> what are the implications to upstream dependencies like rover and router with a change like this?

So in the immeadiate term (i.e. on merging this PR) nothing, it's just another option that's available. `rover` could then start to offer composition on platforms that didn't previously support it as we'd have a `supergraph` binary we could use. For `router` I'm not sure there's any impact at all because all of this only affects the `supergraph` binary, not Harmonizer and the other parts, at least not that I can see?

PR #567, opened 2024-08-23
🐛 fix subgraph location from ast

Jephuff:
We were returning `None` if there was no location, even if we had a subgraph. If there is a subgraph, but no location we still want to return `Some` and just have a `None` range, this way the error can still be mapped to the subgraph instead of being treated as a global.

### DIFF
--- a/apollo-composition/src/lib.rs
+++ b/apollo-composition/src/lib.rs
@@ -287,18 +287,19 @@ impl From<SubgraphLocation> for BuildMessageLocation {
 
 impl SubgraphLocation {
     fn from_ast(node: SubgraphASTNode) -> Option<Self> {
-        let loc = node.loc?;
         Some(Self {
-            subgraph: node.subgraph.unwrap_or_default(),
-            range: Some(Range {
-                start: LineColumn {
-                    line: loc.start_token.line?,
-                    column: loc.start_token.column?,
-                },
-                end: LineColumn {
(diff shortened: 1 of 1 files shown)


trevor-scheer (approved):

apollo-composition/src/lib.rs
```diff
@@ -287,18 +287,19 @@ impl From<SubgraphLocation> for BuildMessageLocation {
...
-                    line: loc.start_token.line?,
-                    column: loc.start_token.column?,
-                },
-                end: LineColumn {
-                    line: loc.end_token.line?,
-                    column: loc.end_token.column?,
-                },
+            subgraph: node.subgraph?,
+            range: node.loc.and_then(|node_loc| {
+                Some(Range {
```
lennyburdette, line 293:
```suggestion
            range: node.loc.map(|node_loc| {
                Range {
```

Jephuff: I think we need `and_then` to handle all the line/column Options below?

lennyburdette: ah sorry i missed that!

lennyburdette (approved):

apollo-composition/src/lib.rs
Jephuff: That `?` operator in rust is pretty unreadable in my opinion :P

PR #568, opened 2024-08-23
Ensure directives are preserved when serializing subgraph schemas in planner

tninesling:


router-bridge/js-src/plan.ts
```diff
@@ -281,7 +282,9 @@ export class BridgeQueryPlanner {
     let result = new Map<string, string>();
 
     subgraphs.names().forEach((name) => {
-      let sdl = printSchema(subgraphs.get(name).schema.toGraphQLJSSchema({}));
+      let sdl = printSchemaWithDirectives(
```
sachindshinde, line 285: You don't have to use `graphql-tools` here, or call `toGraphQLJSSchema()`. There's a `printSchema()` function in `@apollo/federation-internals` (code [here](https://github.com/apollographql/federation/blob/6b52c60d4ee9c77fee832d119d9ef752f51e16e6/internals-js/src/print.ts#L94)), if you import that you can just use `printSchema(subgraphs.get(name).schema)`.

sachindshinde (changes requested):

router-bridge/js-src/plan.ts
tninesling: Updated to use the one from fed internals in [4ff041b](https://github.com/apollographql/federation-rs/pull/568/commits/4ff041b6f9936a9e45b2eabead1194b4b57e7a07)

SimonSapin:
This PR targets the branch `router-bridge@v0.6.0-beta.0+v2.9.0-beta.0` but that release is already out: https://crates.io/crates/router-bridge/0.6.0-beta.0+v2.9.0-beta.0

router-bridge/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "router-bridge"
-version = "0.6.0-beta.0+v2.9.0-beta.0"
+version = "0.6.0-beta.0+v2.9.0-beta.1"
```
sachindshinde, line 3: (partial review, still looking at code) Note that this should be `0.6.0-beta.1+v2.9.0-beta.0`, not `0.6.0-beta.0+v2.9.0-beta.1`.

sachindshinde (changes requested):

router-bridge/Cargo.toml
tninesling: Ok, I can update that. Per @SimonSapin's comment, should I rebase this and target `main`? Or is it ok to release a `beta.1` from this branch?

sachindshinde: I think we've been releasing RCs off branches and not `main` (or at least, recently we have been).

router-bridge/js-src/plan.ts
```diff
@@ -281,7 +282,7 @@ export class BridgeQueryPlanner {
     let result = new Map<string, string>();
 
     subgraphs.names().forEach((name) => {
-      let sdl = printSchema(subgraphs.get(name).schema.toGraphQLJSSchema({}));
+      let sdl = printSchemaWithDirectives(subgraphs.get(name).schema);
```
sachindshinde, line 285: (This comment is more of an FYI.) Looking at [the PR](https://github.com/apollographql/federation-rs/pull/371) that introduced `subgraphs()`, it seems this method is used to generate subgraph schemas to validate subgraph responses against. Two things here:
1. I'm guessing here that the presence of directive applications in the subgraph schema shouldn't impede this subgraph response validation.
2. We're also presumably asserting here that the subgraph schemas with directive applications are valid GraphQL (which they ought to be), but I checked and that's not a new assertion in router (we use the default of `shouldValidate` when calling `Supergraph.build()` in router), so it should be fine (i.e., if there were bugs in how we generated/printed subgraph schemas, they wouldn't suddenly be triggered after this PR).

router-bridge/src/planner.rs
```diff
@@ -2466,4 +2466,26 @@ feature https://specs.apollo.dev/unsupported-feature/v0.1 is for: SECURITY but i
...
+            use std::fmt::Write;
+
+            let sdl = subgraphs.get(name).unwrap();
+            _ = writeln!(
+                &mut snapshot,
+                "--------------------\n{}\n--------------------\n{}",
+                name, sdl
+            );
+        }
+        insta::assert_snapshot!(snapshot);
```
sachindshinde, line 2489: This is more a nit about consistency, but do we want to do something similar to [the other tests around `subgraphs()`](https://github.com/apollographql/federation-rs/pull/371)? E.g.
```rust
        let subgraphs = planner.subgraphs().await.unwrap();
        let subgraphs: BTreeMap<String, String> = subgraphs.into_iter().collect();
        for (name, schema) in subgraphs {
            insta::assert_snapshot!(schema);
        }
```

sachindshinde (changes requested):

router-bridge/src/planner.rs
tninesling: I've been following the pattern used here https://github.com/apollographql/router/blob/dev/apollo-federation/tests/extract_subgraphs.rs#L93

Switching to `BTreeMap` should resolve the ordering issue I was solving by hard-coding the names through

router-bridge/js-src/plan.ts
tninesling: > I'm guessing here that the presence of directive applications in the subgraph schema shouldn't impede this subgraph response validation.

It shouldn't affect that part of the code. One thing I'm noticing when running the router tests against this change is that it changes the query hash in a few tests because the hash will include directives if they exist. So, this will bust the cache, but that happens anyway when we bump the federation version. I think this will still be correct.

router-bridge/Cargo.toml
tninesling: I realized I'm not super familiar with how we release this package. Is the branch name significant, or is it fine to merge here after approval and publish from this branch?

router-bridge/src/testdata/custom_cost_schema.graphql
```diff
@@ -0,0 +1,154 @@
...
+    import: ["@cost", "@listSize"]
+  ) {
+  query: Query
+}
+
+directive @cost(
+  weight: Int!
+) on ARGUMENT_DEFINITION | ENUM | FIELD_DEFINITION | INPUT_FIELD_DEFINITION | OBJECT | SCALAR
+
+directive @cost__listSize(
```
sachindshinde, line 15: I'm confused why there's a directive definition for `@cost__listSize` here since `@listSize` was imported, the directive's name in schema should be `@listSize` instead of `@cost__listSize`.

sachindshinde (changes requested):

router-bridge/src/testdata/custom_cost_schema.graphql
tninesling: It's an artifact of composition when we have multiple directives defined in the same specification. I copied this over from other test cases to make sure it works. We can try to remove that extra definition later, but it's a composition change, so there's nothing I can do in this PR.

sachindshinde: @tninesling 
The `@link` spec generally assumes that directives in a spec have one name in a schema (and code that processes a schema using `@link` can and often do assume this). This is technically fine in the example schema given because the extra definition gets ignored, but applications of this definition may also be ignored (and that's usually not permissible).

You're right that this isn't a problem with this PR specifically, so I'll go ahead and approve, and release the new RC. I'll see if I can get the composition issue fixed tonight (if so, I'll cut a `2.9.0-beta.1` RC).

sachindshinde (approved):

PR #574, opened 2024-09-05
apollo-federation-types: merge routing_url in `SupergraphConfig::merge_subgraphs`

glasser:
This function was added a few months ago in #541 and as far as I know is only used in Rover for the `supergraph compose` and `dev` commands to combine configuration from `--graph-ref` and
`--supergraph-config`/`--config` (and the feature does not actually work for `supergraph compose` due to the issue fixed in https://github.com/apollographql/rover/pull/2101).

The use case here is to let you run composition or dev against a GraphOS graph with some local changes, and so "I just want to change the SDL and nothing else" is a reasonable use case. This change means that if an override specifies SDL but no routing URL for a subgraph, Rover will continue to use the routing URL fetched from GraphOS for that subgraph, which seems reasonable.

While this is a "backwards incompatible" change, this crate is only intended for use by the Rover CLI and in the one case where this was called (`rover dev --graph-ref REF --supergraph-config CONFIG`) this seems like a clear improvement.

### DIFF
--- a/apollo-federation-types/src/config/supergraph.rs
+++ b/apollo-federation-types/src/config/supergraph.rs
@@ -112,10 +112,24 @@ impl SupergraphConfig {
         self.federation_version.clone()
     }
 
-    /// Merges the subgraphs of another [`SupergraphConfig`] into this one
+    /// Merges the subgraphs of another [`SupergraphConfig`] into this one; the
+    /// other config takes precedence when there are overlaps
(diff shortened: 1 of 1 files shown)


benjamn (approved):
Seems reasonable, has a test, and the test fails without the changes.

tninesling (approved):

PR #582, opened 2024-09-13
Add JSONSchema generation support for `supergraph.yaml` files

dylan-apollo:
This was done as a one-off for our extension (which will soon include validation for `supergraph.yaml` files!), but I figure we should keep it around so we can regenerate again in the future.

We can/should also add a command in `rover` to do this, similar to the equivalent of `router`, so folks can generate schemas for their IDEs of choice.

Removes `camino` because it is not supported by `schemars`.

### DIFF
--- a/apollo-federation-types/CHANGELOG.md
+++ b/apollo-federation-types/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 Not every version is listed here because versions before 0.14.0 did not have a changelog.
 
+## 0.15.0 - Unreleased
+
+### Breaking changes
+
+- All usages of `camino::Utf8PathBug` have been replaced with `std::path::PathBuf`
+
+### Features
+
+- A new `json_schema` feature derives the `schemars::JsonSchema` trait on `SupergraphConfig` and its sub-types.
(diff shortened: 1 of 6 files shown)


apollo-federation-types/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 Not every version is listed here because versions before 0.14.0 did not have a changelog.
 
+## 0.15.0 - Unreleased
+
+### Breaking changes
+
+- All usages of `camino::Utf8PathBug` have been replaced with `std::path::PathBuf`
```
lennyburdette, line 9:
```suggestion
- All usages of `camino::Utf8PathBuf` have been replaced with `std::path::PathBuf`
```
🐛 🤣

lennyburdette (dismissed):

apollo-federation-types/CHANGELOG.md
dylan-apollo: You can see how I really feel 😝

dotdat (dismissed):

PR #583, opened 2024-09-17
fix: remove connectors warning

lennyburdette:


apollo-composition/src/lib.rs
```diff
@@ -133,23 +134,27 @@ pub trait HybridComposition {
...
+                        for (service_name, connector) in by_service_name.iter() {
+                            issue.message = issue
+                                .message
+                                .replace(&**service_name, connector.id.subgraph_name.as_str());
+                        }
+                        issue
+                    }),
+                );
+
+                if env::var("CONNECTORS_PREVIEW").unwrap_or_default() != "true" {
```
nicholascioli, line 147: Small nit, but do we want users to be able to use this env var like others? If so, we need to handle cases where it is equal to other truthy things (e.g. `1`, `"1"`, `TRUE`, etc.). The router gets this functionality by using clap's [`FalseyValueParser`](https://docs.rs/clap/latest/clap/builder/struct.FalseyValueParser.html).

nicholascioli (dismissed):
Looks good to me! I left a small nit about truthy env vars, but feel free to ignore.

apollo-composition/src/lib.rs
lennyburdette: i'm really just trying to avoid some misleading stuff in demos. i actually can't figure out how to set this env var in the appropriate place. i might just make a custom build for demos that removes this

nicholascioli: Super fair. Feel free to wrap this in maybe a `#[cfg(debug_assertions)]` or a custom feature if you want!

dylan-apollo: I think if this is controlled via env var, we probably want to set this on the supergraph binary itself. Then we can use the same Clap options as we do for other env vars

dylan-apollo: That would mean making it a Rust option here—which could also be set in the web editor if we want.

Do we just want to get rid of or change this warning, though?

lennyburdette: yeah maybe. warning logs in router might be better now that composition errors show up in more tools

dylan-apollo (approved):

PR #590, opened 2024-10-03
Fix deserialization of GraphQLError nodes

esilverm:
While attempting to deserialize certain composition errors, we ran into some errors which caused crashes. After some digging it was determined that it was due to the `nodes` key on `GraphQLError` being set as `undefined` in our JSON response.

In the fed-types crate, we attempt to use `#[serde(default)]` to handle this case, however as mentioned in #563, it only works when they key is not present. Instead we should handle it as an `Option` to account for its absence or rare `undefined` assignment.

### DIFF
--- a/apollo-composition/CHANGELOG.md
+++ b/apollo-composition/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- [#590](https://github.com/apollographql/federation-rs/pull/590) Fix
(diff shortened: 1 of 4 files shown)


dylan-apollo:
@esilverm can you add notes to the CHANGELOG for `apollo-composition` (bug fix/patch) and `apollo-federation-types` (breaking change)?

dylan-apollo (approved):
Perfect, thanks!

PR #591, opened 2024-10-10
Use LatestFedTwo as default federation version

dotdat:


dylan-apollo:
I think this would end up being a breaking change for customers, where if they don't have a specific federation version set in their `supergraph.yaml`, we're suddenly going to start upgrading them to fed 2 mode, right?

dariuszkuc:
Wondering whether latest is the right default, i.e. we definitely want to update the default to 2.x but I'd argue that we probably would want to target something like latest-1 (or some fixed fed version) for stability purposes.

Since we were defaulting to fed 1, customers wouldn't be using any fed 2 features so technically we could point to just to fed 2.0.

abernix:
Without much consideration, I'm just going to close this.  I think it's abandoned. :)   Feel free to reopen.

PR #594, opened 2024-11-07
bump apollo-federation to 2.0.0-preview.1

lennyburdette:


dylan-apollo (approved):
LGTM. There is also a pending release of `apollo-federation-types`, but I _think_ we can wait on that for another time? If cargo publish fails, though, that might be why

dylan-apollo (approved):

PR #602, opened 2024-12-19
composition: add validation for @override

nicholascioli:
The `@override` directive does not work with connector-related subgraphs due to internal implementation details, so this commit adds validation rules to error if any `@override` directive tries to override fields from connector-aware subgraphs.

Since there are breaking changes in connector-related code in `apollo-federation`, this commit also pins the version of upstream to a specific git revision.


nicholascioli:
This is blocked until #603 is merged

apollo-composition/src/lib.rs
```diff
@@ -101,6 +116,87 @@ pub trait HybridComposition {
...
+                    ExtendedType::Object(node) => extract_directives!(node),
+                    ExtendedType::Interface(node) => extract_directives!(node),
+                    ExtendedType::InputObject(node) => extract_directives!(node),
+
+                    // These types do not have fields
+                    ExtendedType::Scalar(_) | ExtendedType::Union(_) | ExtendedType::Enum(_) => {
+                        Vec::new()
+                    }
+                })
+                .filter(|(_, directive)| directive.name == "override");
```
lennyburdette, line 157: unfortunately, we don't know if the directive name is `@override`, `@federation__override`, or some other alias the user chose.

let's check for `directive.name == "override" || directive.name == "federation__override"` and put a TODO for handling aliases later.

apollo-composition/src/lib.rs
```diff
@@ -101,6 +116,87 @@ pub trait HybridComposition {
...
+                };
+
+                if let Some(overriden_subgraph) =
+                    validations_by_subgraph_name.get(overriden_subgraph_name)
+                {
+                    if overriden_subgraph.has_connectors {
+                        override_errors.push(Issue {
+                            code: "OVERRIDE_ON_CONNECTOR".to_string(),
+                            message: format!(
+                                r#"Field "{}" on subgraph "{}" is trying to override connector-enabled subgraph "{}", which is not yet supported"#,
```
lennyburdette, line 177:
```suggestion
                                r#"Field "{}" on subgraph "{}" is trying to override connector-enabled subgraph "{}", which is not yet supported. See https://go.apollo.dev/connectors/limitations#override-is-partially-unsupported"#,
```

apollo-composition/src/lib.rs
```diff
@@ -101,6 +116,87 @@ pub trait HybridComposition {
             return;
         };
 
+        // Overrides mess with the supergraph in ways that can be difficult to detect when
+        // expanding connectors; the supergraph may omit overridden fields and other shenanigans.
+        // To allow for a better developer experience, we check here if any connector-enabled subgraphs
+        // have fields overridden and error early.
+        let validations_by_subgraph_name =
```
lennyburdette, line 123: whaddaya think about put think in its own function just for readability?

lennyburdette (approved):

PR #603, opened 2024-12-19
remove router-bridge

nicholascioli:
Now that the router has updated its MSRV to a version which is not supported by deno, keeping `router-bridge` around breaks the release process for `apollo-composition`. This commit removes it entirely while also bumping the version of rust used by circleci to the latest stable version.


lennyburdette (approved):
🥳

lennyburdette (approved):

PR #613, opened 2025-01-23
prepend [subgraph] to issues

lennyburdette:
both rover and graphos are currently unable to use the location field to display errors. and even if they did, js federation usually doesn't include location data so this is the best we can do for now

### DIFF
--- a/apollo-federation-types/src/composition/mod.rs
+++ b/apollo-federation-types/src/composition/mod.rs
@@ -1,5 +1,6 @@
 //! Types used with the `apollo-composition` crate
 
+use std::collections::HashSet;
 use std::ops::Range;
 
 use apollo_compiler::parser::LineColumn;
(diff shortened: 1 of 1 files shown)


apollo-federation-types/src/composition/mod.rs
```diff
@@ -88,11 +89,35 @@ impl From<BuildHint> for Issue {
...
+/// the subgraph name if there's exactly one. If there's more than one, it's
+/// probably a composition issue that's not attributable to a single subgraph,
+/// and GraphOS will show "[subgraph unknown]", which is also not correct.
+fn maybe_prepend_subgraph(message: &str, locations: &[SubgraphLocation]) -> String {
+    if message.starts_with('[') {
+        return message.to_string();
+    }
+    let unique_subgraphs = locations
+        .iter()
+        .filter_map(|l| l.subgraph.clone())
```
pubmodmatt, line 103: Nitpick, but you can avoid cloning every item:
```suggestion
        .filter_map(|l| l.subgraph.as_ref())
```

apollo-federation-types/src/composition/mod.rs
```diff
@@ -88,11 +89,35 @@ impl From<BuildHint> for Issue {
     }
 }
 
+/// Rover and GraphOS expect messages to start with `[subgraph name]`. (They
+/// don't actually look at the `locations` field, sadly). This will prepend
+/// the subgraph name if there's exactly one. If there's more than one, it's
+/// probably a composition issue that's not attributable to a single subgraph,
+/// and GraphOS will show "[subgraph unknown]", which is also not correct.
+fn maybe_prepend_subgraph(message: &str, locations: &[SubgraphLocation]) -> String {
```
pubmodmatt, line 97: Could you add  a quick unit test?

lennyburdette: it gets tested in fedutil-build-plugins but 👍🏻 for finding problems sooner

PR #621, opened 2025-03-04
Add RouterVersion::LatestTwo

pubmodmatt:
* Rename `RouterVersion::Latest` to `RouterVersion::LatestOne`
* Add `RouterVersion::LatestTwo`
* Bump version numbers of `apollo-federation-types` and `apollo-composition` for release

### DIFF
--- a/apollo-federation-types/src/config/version.rs
+++ b/apollo-federation-types/src/config/version.rs
@@ -76,8 +76,8 @@ impl FromStr for RouterVersion {
             }
         } else {
             match input {
-                "1" | "latest" => Ok(Self::LatestOne),
-                "2" => Ok(Self::LatestTwo),
+                "1" => Ok(Self::LatestOne),
+                "2" | "latest" => Ok(Self::LatestTwo),
                 _ => Err(invalid_version),
             }
         }


apollo-federation-types/src/config/version.rs
```diff
@@ -72,7 +76,8 @@ impl FromStr for RouterVersion {
             }
         } else {
             match input {
-                "1" | "latest" => Ok(Self::Latest),
+                "1" => Ok(Self::LatestOne),
+                "2" | "latest" => Ok(Self::LatestTwo),
```
pubmodmatt, line 80: @jonathanrainer - is it correct to update `latest` here?

pubmodmatt: Actually I'm thinking it's too soon for this - if some used `latest` and it used `latest-2` without Orbiter being up to date, that could cause problems, right?

pubmodmatt: Had it right ht efirst time, `latest` should evaluate to `LatestTwo`

jonathanrainer (approved):
LGTM

dotdat (approved):

PR #630, opened 2025-05-16
expose Rust native composition steps

dariuszkuc:
Updates hybrid composer to expose more granular composition steps. All new methods are prefixed with `experimental_`.


dariuszkuc:
we won't be able to merge it until the next `apollo-federation` crate release

apollo-composition/src/lib.rs
```diff
@@ -166,6 +142,291 @@ pub trait HybridComposition {
...
+    /// 8. Post merge validations
+    /// 9. <connectors> expand supergraph
+    /// 10. Validate satisfiability
+    ///
+    /// In case of a composition failure, we return a list of errors from the current composition
+    /// phase.
+    async fn experimental_compose(
+        &mut self,
+        subgraph_definitions: Vec<SubgraphDefinition>,
+    ) -> Result<CompositionResult, Vec<Issue>> {
```
dariuszkuc, line 167: this actually should be `Result<PluginResult, Vec<issues>>`

apollo-composition/src/lib.rs
```diff
@@ -166,6 +142,291 @@ pub trait HybridComposition {
...
+                    })
+            })
+            .collect::<Vec<_>>();
+
+        let errors: Vec<Issue> = subgraph_validation_errors
+            .iter()
+            .filter(|issue| issue.severity == Severity::Error)
+            .cloned()
+            .collect();
+        self.add_issues(subgraph_validation_errors.into_iter());
```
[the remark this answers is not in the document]
dariuszkuc: this method is called from old code (that relies on self.issues) and new code (that returns issues directly)

apollo-federation-types/src/composition/mod.rs
```diff
@@ -89,6 +90,43 @@ impl From<BuildHint> for Issue {
     }
 }
 
+impl From<FederationError> for Issue {
```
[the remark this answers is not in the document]
dariuszkuc: yes

apollo-federation-types/src/javascript/mod.rs
```diff
@@ -27,6 +29,12 @@ pub struct SatisfiabilityResult {
     pub hints: Option<Vec<CompositionHint>>,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
+pub struct CompositionResult {
+    pub supergraph: String,
```
clenfest, line 34: Can this not also return CompositionErrors and fail to compose?

apollo-federation-types/src/javascript/mod.rs
```diff
@@ -69,3 +77,52 @@ pub struct GraphQLError {
...
+
+impl TryFrom<SubgraphDefinition> for Subgraph<Initial> {
+    type Error = SubgraphError;
+
+    fn try_from(value: SubgraphDefinition) -> Result<Self, Self::Error> {
+        Subgraph::parse(value.name.as_str(), value.url.as_str(), value.sdl.as_str())
+    }
+}
+
+impl TryFrom<SubgraphDefinition> for Subgraph<Upgraded> {
```
dariuszkuc, line 89: instead of try from trait move it to a separate method to make it explicit that we are ignoring validation

apollo-federation-types/src/composition/mod.rs
```diff
@@ -89,6 +90,43 @@ impl From<BuildHint> for Issue {
...
+            // TODO CompositionError should specify locations
+            locations: vec![],
+            severity: Severity::Error,
+        }
+    }
+}
+
+impl From<SubgraphError> for Issue {
+    fn from(error: SubgraphError) -> Self {
+        let (code, message) = match error.format_errors().into_iter().next() {
```
dariuszkuc, line 114: add extra validation that we only have a single error

apollo-federation-types/src/javascript/mod.rs
dariuszkuc: Yeah, I've updated `experimental_composition` to explicitly return hints on success and errors for satisfiability failures

apollo-composition/src/lib.rs
```diff
@@ -172,13 +145,305 @@ pub trait HybridComposition {
...
+    /// 8. Post merge validations
+    /// 9. <connectors> expand supergraph
+    /// 10. Validate satisfiability
+    ///
+    /// In case of a composition failure, we return a list of errors from the current composition
+    /// phase.
+    async fn experimental_compose(
+        mut self,
+        subgraph_definitions: Vec<SubgraphDefinition>,
+    ) -> Result<PluginResult, Vec<Issue>>
```
dylan-apollo, line 170: I doubt we want to use `PluginResult`, it's kind of a clunky type that we had to work around a bunch in fbp. The only reason we're still using it at all is for compatibility with `bps`, which is 1 of 3 consumers of this function, ultimately.

All we should really need is effectively `Result<(Schema, Vec<Issue>), Vec<Issue>>`, but a proper type instead of the tuple, right?

apollo-composition/src/lib.rs
```diff
@@ -172,13 +145,305 @@ pub trait HybridComposition {
...
+            Ok(results) => results,
+            Err(errors) => {
+                return Err(errors);
+            }
+        };
+        let upgraded_subgraphs = self.experimental_upgrade_subgraphs(subgraphs).await?;
+        let validated_subgraphs = self
+            .experimental_validate_subgraphs(upgraded_subgraphs)
+            .await?;
+
```
dylan-apollo, line 189: We probably want to actually run Connectors validations after other subgraph validations now. The only reason it was first originally was because the JS steps are sooo slow. We probably want to operate on the real (not upgraded by connectors) schema as long as possible, so up until overrides & merge.

dylan-apollo (commented):
Can you also add a changelog entry for this crate?

apollo-composition/src/lib.rs
```diff
@@ -1,17 +1,26 @@
-use std::collections::HashMap;
-use std::iter::once;
-
 use apollo_compiler::{schema::ExtendedType, Schema};
+use apollo_federation::composition::{
+    expand_subgraphs, merge_subgraphs, post_merge_validations, pre_merge_validations,
```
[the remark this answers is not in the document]
dariuszkuc: I'd imagine that `expand_connectors` should be called from general `expand_subgraphs` but yeah we can try to come up with some better names.

apollo-composition/src/lib.rs
```diff
@@ -172,13 +145,305 @@ pub trait HybridComposition {
...
+    /// 6. Pre-merge validations (includes connectors validations)
+    /// 7. Merge subgraphs into a supergrpah
+    /// 8. Post merge validations
+    /// 9. <connectors> expand supergraph
+    /// 10. Validate satisfiability
+    ///
+    /// In case of a composition failure, we return a list of errors from the current composition
+    /// phase.
+    async fn experimental_compose(
+        mut self,
```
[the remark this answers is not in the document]
dariuszkuc: Actually we do need this. We are calling all other `experimental_` methods which require `&mut self` so we can call JS through deno. The intent here was to have the same trait implemented by legacy and native composers.

apollo-composition/src/lib.rs
```diff
@@ -172,13 +145,305 @@ pub trait HybridComposition {
...
+        let merge_result = self
+            .experimental_merge_subgraphs(validated_subgraphs)
+            .await?;
+
+        // expand connectors as needed
+        let supergraph_sdl = merge_result.supergraph.clone();
+        let expansion_result = match expand_connectors(&supergraph_sdl, &Default::default()) {
+            Ok(result) => result,
+            Err(err) => {
+                return Err(vec![err.into()]);
```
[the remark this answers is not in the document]
dariuszkuc: If merge fails this will not get executed. In JS code we only return hints if satisfiability succeeds.

TylerBloom (approved):

PR #640, opened 2025-07-15
Add support of cacheTag directive validation

bnjjj:
Waiting for release of `apollo-federation` crate.


duckki (commented):
I pushed a minor fix. Otherwise, it looks good!

PR #641, opened 2025-07-15
deps: update schemars to v1.0.4

DaleSeo:
Following PR #637 where `schemars` was upgraded from v.0.8.21 to v1.0.0, this PR updates it to v1.0.4 which is the latest version. This is currently blocking Rover from updating in https://github.com/apollographql/rover/pull/2641.

### DIFF
--- a/apollo-federation-types/Cargo.toml
+++ b/apollo-federation-types/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT"
 name = "apollo-federation-types"
 readme = "README.md"
 repository = "https://github.com/apollographql/federation-rs/"
-version = "0.15.7"
+version = "0.15.8"
 
 [features]
 default = ["config", "build", "build_plugin"]
--- a/apollo-federation-types/Cargo.toml
+++ b/apollo-federation-types/Cargo.toml
@@ -26,7 +26,7 @@ apollo-federation = { workspace = true }
 
 # config and build dependencies
 serde = { version = "1", features = ["derive"] }
-schemars = { version = "1.0.0", optional = true, features = ["url2"] }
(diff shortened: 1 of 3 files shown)


apollo-federation-types/Cargo.toml
```diff
@@ -26,7 +26,7 @@ apollo-federation = { workspace = true }
 
 # config and build dependencies
 serde = { version = "1", features = ["derive"] }
-schemars = { version = "1.0.0", optional = true, features = ["url2"] }
+schemars = { version = "1.0.4", optional = true, features = ["url2"] }
```
duckki, line 29: Do we really need to pin it down to the patch version? Why not `"1"` or `"1.0"` here?

DaleSeo: Hey @duckki, I didn't mean it that way. Sorry, I'm new to Cargo 😅. I've changed it to "1" now. Thanks for the feedback! 🙏

duckki: It was a genuine question. So, I don't know which is the right one. In general, I try not to pin too specifically, but there are exceptions.

goto-bus-stop: Specifying 1.0.4 means "latest 1.x version, but at least 1.0.4", so it's not really that different from specifying `1` or `1.0.0`. So it's not _pinned_ unless you write `=1.0.4`. In most cases it's fine to just specify the most recent available version, if you're working on a frequently used library then you might want to be more careful about actually specifying the minimum version you support.

goto-bus-stop: yeah, both are totally fine here 😁

PR #648, opened 2025-08-05
BREAKING CHANGE: update `HybridComposition` methods to use `Issue` type for hints

duckki:
### Background

We are anticipating our new Rust composition will be a drop-in replacement of HybridComposer. But, currently, `HybridComposition` trait is using the JS-specific `CompositionHint` type for hints (in both `MergeResult` and `SatisfiabilityResult`).

The problem is the Rust composition has to emulate the JS data structures as it currently stands. It will be much easier if the trait methods use the `Issue` type in the first place.

### Proposed Changes

This PR changes `HybridComposition` trait to use the `Issue` type directly, instead of JS `CompositionHint`. This is intended to be a pure refactoring. `CompositionHint` values from `HybridComposition` are immediately converted to `Issue` type anyways. We can use `Issue` type for API and do the conversion in the fedutils-build-plugin's implementation with no effective changes.

* changed `MergeResult` to use `Issue` type.
  * This is only used for `experimental_` methods. So, this is not a breaking change.
* moved `MergeResult` from `javascript` module to `composition` module, since it does not concern JS serialization.
* changed `validate_satisfiability` trait method to return `Vec<Issue>`, instead of `SatisfiabilityResult`.
  * This is technically a breaking change, but it shouldn't be used outside of this apollo-composition & fbp and not visible to customers.
  * If we want to keep the API in the main branch, we can target `next` branch, instead.
  * Note: The method previously returned `Result<SatisfiabilityResult, Issue>`, which really had 3 cases -- success with optional JS hints, satisfiability error with a list of JS errors, or deno error with one `Issue`. This PR conflates the latter two as `Err(Vec<Issue>)`. I think that's ok since the downstream usage of this method does not really distinguish those cases.

Companion PR on fbp repo: https://github.com/mdg-private/fedutils-build-plugins/pull/151

### DIFF
--- a/apollo-composition/src/lib.rs
+++ b/apollo-composition/src/lib.rs
@@ -10,8 +10,10 @@ use apollo_federation::connectors::{
 };
 use apollo_federation::subgraph::typestate::{Initial, Subgraph, Upgraded, Validated};
 use apollo_federation::subgraph::SubgraphError;
-use apollo_federation_types::build_plugin::{MergeResult, PluginResult};
-use apollo_federation_types::composition::{convert_subraph_error_to_issues, SubgraphLocation};
+use apollo_federation_types::build_plugin::PluginResult;
+use apollo_federation_types::composition::{
+    convert_subraph_error_to_issues, MergeResult, SubgraphLocation,
+};
 use apollo_federation_types::{
     composition::{Issue, Severity},
     javascript::SubgraphDefinition,
--- a/apollo-composition/src/lib.rs
+++ b/apollo-composition/src/lib.rs
@@ -44,8 +44,8 @@ pub trait HybridComposition {
     ///
     /// # Output
     ///
-    /// If satisfiability completes from JavaScript, either a list of hints (the Ok case) or a list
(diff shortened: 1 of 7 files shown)


apollo-composition/src/lib.rs
```diff
@@ -161,12 +159,13 @@ pub trait HybridComposition {
...
     /// 10. Validate satisfiability
     ///
+    /// If successful, returns a tuple of (the supergraph SDL, a list of hints).
     /// In case of a composition failure, we return a list of errors from the current composition
     /// phase.
     async fn experimental_compose(
         mut self,
         subgraph_definitions: Vec<SubgraphDefinition>,
-    ) -> Result<PluginResult, Vec<Issue>>
+    ) -> Result<(String, Vec<Issue>), Vec<Issue>>
```
dariuszkuc, line 168: IMHO returning proper type is clearer than a tuple, i.e. what is the meaning behind `String` and `Vec<Issue>` that are being returned?

dariuszkuc (dismissed):

duckki:
I created a new PR targeting the `next` branch: https://github.com/apollographql/federation-rs/pull/651
Closing.

PR #693, opened 2026-04-29
chore: update to apollo-federation/router 2.14.0

SharkBaitDLS:
I also changed it so that the apollo-federation crate is pinned to its exact version, which will prevent Rover from transitively pulling in newer apollo-federation versions early before this crate is updated to work with them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,7 +57,7 @@ dependencies = [
 
 [[package]]
 name = "apollo-composition"
-version = "0.5.2"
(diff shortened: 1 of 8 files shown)


briannafugate408 (approved):

duckki:
@briannafugate408 @SharkBaitDLS  FYI, we could merge the `dev` branch into `main`, since the `dev` branch already had the update. But, this is not a big drift. So, not a big deal.

PR #695, opened 2026-04-30
feat: skip redundant serialization in composition

dariuszkuc:
We can avoid unnecessary serialization and deserialization of upgraded+validated subgraph.

### DIFF
--- a/apollo-composition/src/lib.rs
+++ b/apollo-composition/src/lib.rs
@@ -1,13 +1,15 @@
 use apollo_compiler::{schema::ExtendedType, Schema};
 use apollo_federation::composition::{
     expand_subgraphs, merge_subgraphs, post_merge_validations, pre_merge_validations,
-    upgrade_subgraphs_if_necessary, validate_satisfiability, Supergraph,
+    upgrade_subgraphs_if_necessary, validate_satisfiability, CompositionFailure, Merged,
+    Supergraph,
 };
 use apollo_federation::connectors::{
     expand::{expand_connectors, Connectors, ExpansionResult},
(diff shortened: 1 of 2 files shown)


apollo-composition/src/lib.rs
```diff
@@ -175,9 +177,6 @@ pub trait HybridComposition {
     where
         Self: Sized,
     {
-        // `@cacheTag` directive validation
-        validate_cache_tag_in_subgraphs(&subgraph_definitions)?;
```
dariuszkuc, line 179: this is already validated as part of regular subgraph federation validations

duckki: Why don't we defer this removal after Rust composition GA to avoid deviation from JS behavior?
Though it is true that this is done a part of Rust subgraph validation, removing this line may change observed behavior of Rust composition when subgraphs have cacheTag validation errors along with other errors.

duckki (approved):
One minor concern regarding cache tag error short-circuiting behavior change. Otherwise, it looks good.

apollo-composition/src/lib.rs
dariuszkuc: While it does deviate from current behavior, why would we want to continue to run it twice? Same validations will run so the same issues will be reported. The difference would be that in JS `@cacheTag` issues would be reported first and no other validations would run vs we would run whole validations in RS and then report all issues at once (which would include `@cacheTag` errors if any). 

IMHO giving users all validations at once is preferable over failing once for `@cacheTag` and then failing second time due to the additional issues.

